PR #52, opened 2020-05-31
Add Content Type Model (Get, Query, Add)

ypcode:
There is still an issue with Add 
The returned created Id mismatches the specified Id


ypcode:
Tests to check equality of Content Type ID has been commented out. The API does not return the specified Id. May be necessary to use Client.svc instead for Content Type creation

ypcode:
Hi guys , I noted the tests failed here... however all the tests are passing with mock data on my side now, not sure why it fails here...

Issue #47, opened 2020-05-26
Extend the SharePoint model for Content Types and Fields


jansenbe:
## Category

- [x] Domain model extension

## Describe the domain model extension

Extending the SharePoint model to work for Content Types and Fields

jansenbe:
Thanks @ypcode for the content type updates. I'll create a new issue to track the need for CSOM to create content types with a given ID (or alternatively we need to inspect if Graph can do this somehow)

ypcode:
Was also my second guess... that if Graph handles it better we might not need the CSOM mimicking altogether for that purpose. But that needs to be checked :)

jansenbe:
Done now, thanks a ton @ypcode !

PR #60, opened 2020-06-08
Adds CRUD for web fields (only FieldText tested)

ypcode:
#47


jansenbe:
Thanks @ypcode, again an awesome job done! I'll try to process this PR later today, I'll contact you over Teams to discuss in case that would be needed.

Issue #65, opened 2020-06-16
Docs - Setup for tests


pkbullock:
## Category

- [x] Bug

You may need a new category for doc updates maybe?

When setting up the unit tests the solution relies on an Azure AD app "PnP Office 365 Management Shell" which needs to either be added or consented. I used PnP PowerShell docs example 11 in https://docs.microsoft.com/en-us/powershell/module/sharepoint-pnp/connect-pnponline?view=sharepoint-ps to set this up.

```powershell
Connect-PnPOnline -Url https://contoso.sharepoint.com -PnPO365ManagementShell
```
All tests build after this.

You ok for me to update the docs?

jansenbe:
Thanks for this one @pkbullock , just merged your PR

PR #66, opened 2020-06-16
Docs Tweak

pkbullock:
* Added step when I setup unit tests to ensure the multi-tenant app is setup correctly
* Me being picky over spelling. ;-)

Reference issue #65


jansenbe:
Thanks a ton @pkbullock. Good catch on the management shell app setup as without that it's hard to run the unit tests as live tests.

PR #71, opened 2020-06-23
Extend SP Field model (add operation for each field type)

ypcode:
Still missing
- AddAsXml
- Taxonomy field
- RatingScale field ( if needed )


ypcode:
squashed everything into a single commit, added the sync add specific methods

jansenbe:
That was an awesome PR @ypcode ! Provides a very complete and easy to use manner to create fields!

PR #76, opened 2020-06-29
Adds support for custom access token provider

ypcode:


jansenbe:
Thanks @ypcode , good additions to enable .Net Blazor support

PR #99, opened 2020-07-31
Cleaned up Teams implementation

JarbasHorst:
- Separated enums in different files
- Created new enum for ChatMessageReactionType
- Removed unused using directives


jansenbe:
Nice work @JarbasHorst! Thanks for you contribution.

PR #100, opened 2020-08-02
Cleaned up ODataQuery implementation

JarbasHorst:
- Separated enums in different files
- Separated models in different files


jansenbe:
Thanks @JarbasHorst !

PR #108, opened 2020-08-23
Support creation of content type with specific Id using CSOM #54

### LINKED ISSUE #54: Support content type creation with a given parent ID
## Category

- [x] Feature request

## Describe the feature

Content type creation via SharePoint REST does not allow you to create a content type with a given (parent) id...CSOM can do this but check (upcoming) Graph options as well

ypcode:
The approach to populate metadata (Rest ID) is debatable, but that makes the existing unit test working mostly untouched.

Closes #54


jansenbe:
Thanks @ypcode , I'll have a look.

Issue #117, opened 2020-08-26
Writing tests for fill in the gaps in the code coverage


pkbullock:
## Category

- [x] Feature request

## Describe the feature

Improving test coverage for project. Currently at 72%. I've started this work, please assign to me. I will avoid the teams to avoid PR clash with issue #95.

Also, full coverage cannot be achieved without running the test on OSX, Linux as well.

I will likely drop in multiple PRs over time as there is quite a bit to cover.

pkbullock:
Raising this, as the tests will have a dependency on a certificate for testing, so need a thumbprint for the settings. @jansenbe would this break automation?

jansenbe:
@pkbullock : Thanks for taking up this task. Let's push test coverage beyond 80%! 

Some comments:

- The GitHub workflow that runs the tests today does run them on Linux, we could use a similar approach to also run the test suite on macOS (see https://docs.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners)
- Not clear where the certificate comes into the mix? Guess a certificate deployed as PFX file with a password can work across environments, not up to date on the cross platform features of the cert store .Net API's
- Also feel free to be more strict in test requirements - docs on how to write tests are pretty basic, think we can be more prescriptive 
- As always, happy to get into a meeting/brainstorm session to discuss about things

jansenbe:
Also note that you can use `TestCommon.RunningInGitHubWorkflow()` to know when you're running in a GitHub Workflow environment. I today use it to skip 2 tests that can only operate as live test and in test automation in GitHub we only want to run offline tests to avoid having an impact on live environments being ready and available

pkbullock:
Cool thanks for the tip, ill disable the tests with the certificate requirement on GitHub.

For the .net on mobile, I hadn't considered these platforms, there is this service https://visualstudio.microsoft.com/app-center/ which does tests on these platforms, its a costed service though. There maybe other services that allow this type of testing. I've not used it before so would need some research.

jansenbe:
Initial focus would be to increase the test coverage as that will help ensure quality going forward. In parallel we can indeed research option to test on other platforms.

jansenbe:
Think we can close this one @pkbullock as we're above 80% test coverage. Let's keep a continued focus on test coverage going forward, feel free to raise concerns if you see new code being added without the needed tests.

PR #121, opened 2020-08-31
More test coverage

pkbullock:
More tests - another block done

- Field Collection Tests
- Bug Fix for missing non-async interface for calculated field

Coverage @ 76%

More PRs to come related to issue #117


jansenbe:
More test goodnesss from @pkbullock . Thanks!

Issue #64, opened 2020-06-16
Review and complete the Web/Site domain model 


jansenbe:
## Category

- [x] Feature request

## Describe the feature

Review and complete the Web/Site domain model : current web/site model was copied from pnp provisioning engine model and still lacks some of the web/site properties. This is a nice and simple change to start with for new contributors. See https://aka.ms/pnp/coresdk/contributors on how to get started.

ypcode:
Temporarily self assigned to implement some of the properties of the Web model

jansenbe:
Closing this one for now, think we cover more than enough for preview 3. Let's create new issues for the individual missing things once they popup. Thanks for all the great work here @ypcode !

PR #128, opened 2020-09-07
Implement simple Web properties (A-D)

ypcode:
Implement a bunch of simple properties of the Web model (from A to D)
Some properties are not present in CSOM, or at least in the documentation, thus cannot be sure if the are read/write or read-only
linked to #64


jansenbe:
Thanks @ypcode !!

PR #136, opened 2020-09-12
Added all simple properties on Site entity #64

ypcode:
This one adds test to cover all properties of the Site entity.
It pointed out that some properties are not loaded (probably since Graph is not called)
linked to #64


jansenbe:
Thanks @ypcode !

Issue #154, opened 2020-09-23
PnP docs site upgrade


pkbullock:
## Category

- [x] Feature request

## Describe the feature

- Alignment of the website to the standard of the other PnP Sites
- Incorporation of the new branding logos
- Ensure accessible/responsive

On initial investigation, the other sites use a mkdocs/material ui skin, have seen a docfx material ui skin, will see how much would to align both.

pkbullock:
pls assign to me

jansenbe:
Thanks @pkbullock , nice work!

PR #159, opened 2020-09-25
New DocFx Theme

pkbullock:
PR includes

- New site theme based on material UI much like the other sites
- Tracking image
- Red Theme
- Accessibility at 92%
- SEO at 99%
- New Logo
- Better footer with credit to theme material creator

Reference issue #154


jansenbe:
We  now have a nice looking site! Thanks a ton @pkbullock !!

PR #163, opened 2020-09-27
Azure Function v3 sample

ypcode:


jansenbe:
Nice work @ypcode !! With this sample in now have a really nice set of samples showing how one can use the PnP Core SDK.

Ofer-Gal:
I tring  the Azure Function v3 Sample from [Here](https://github.com/pnp/pnpcore/tree/dev/samples/Demo.AzureFunction) but my Target is .Net 5 and the VS created project uses Program.cs to start instead of your  Startup : FunctionsStartup 
Should I just remove program.cs and use your Startup.cs?
Thank you

PR #197, opened 2020-10-16
Update README.md

runeemde:


jansenbe:
Nice catch @runeemde ! Thanks for creating a PR to fix it.

PR #251, opened 2020-12-09
Fixed project references for Az Function demo

s-KaiNet:
Project references were broken for the Azure Functions demo. They were restored to point to the right projects. Also updated dependency `Microsoft.Identity.Client` to version `4.21` to match PnP.Core.


jansenbe:
Good catch @s-KaiNet , the other samples use the nuget releases but this one seems to still use a direct reference.

PR #302, opened 2021-01-21
Add some enhancements

sebastianmattar:
- Add additional configuration options (RedirectUri etc)
- Do not enforce resource parameter for GetAccessTokenAsync(), would break authentication with the Graph API


jansenbe:
Hi @sebastianmattar ,

Thanks for this massive PR. I would have been better to maybe create separate PR's for the three fixes as we're not (yet) convinced that we should include a specific ACS handler given you already have the option to either use the [ExternalAuthenticationProvider](https://pnp.github.io/pnpcore/api/PnP.Core.Auth.ExternalAuthenticationProvider.html) that allows you to provide an access token of your choice or you can also write your own [custom authentication provider](https://pnp.github.io/pnpcore/using-the-sdk/custom-authentication-provider.html) all together. Using Azure ACS is, although still supported, not a highly recommended approach, hence we're hesitant to add this to PnP Core SDK. We'd like to standardize on what can be done with MSAL in PnP.Core.Auth.

Would it be possible to split this PR and have one with the non ACS related changes as well?

jansenbe:
Also, we're open for an improved (e.g. thread safe) ACS implementation in PnP Framework, just not in PnP Core SDK. Hope this feedback and the feedback in the PnP Framework discussion clarifies things.

sebastianmattar:
Alright, I will rework the PR, move the ACS bits to a new PR in PnP.Framework tomorrow (thread-safe and all :).

sebastianmattar:
I removed the ACS bits

PaoloPia:
Hi @sebastianmattar,
Thanks for your great contribution. I've started reviewing this PR and I will keep you posted. Great job!

PaoloPia:
Hi @sebastianmattar,
Could you please expand a little bit more about the reason why you don't want to specify the resource when consuming Microsoft Graph? Thanks.

sebastianmattar:
Yes... I found that in some scenarios only a scope parameter is provided.
Example: [https://github.com/pnp/pnpframework/blob/dev/src/lib/PnP.Framework/AuthenticationManager.cs#L438](https://github.com/pnp/pnpframework/blob/dev/src/lib/PnP.Framework/AuthenticationManager.cs#L438)
Is this a valid token request? How should we handle it?

PaoloPia:
Hi @sebastianmattar,
Your PR is now reviewed and merged. You can freely remove your local branch. Thanks.

PaoloPia:
About your comment regarding the missing resource Uri, that's related to PnP Framework, while in PnP Core we need the resource URI. As such, I reverted back that part of the changes, together with some other fixes and changes needed to successfully execute the automated tests of the PnP.Core.Auth library. Thanks.

Issue #335, opened 2021-02-23
ListItem: Add RoleAssignment collection


jimmywim:
## Category

- [x] Domain model extension

## Describe the domain model extension

Get ListItemAllFields.RoleAssignments

### Model

Model: ListItem
Properties: RoleAssignments

### APIs

API : https://<tenant>.sharepoint.com/sites/<site>/_api/web/lists(guid'<listId>')/item(<ItemId>)/roleAssignments

### Additional Information

I've made a start on this - but I'd appreciate some help. 

I've wired up the `IRoleAssignmentCollection` property to `IListItem` but my unit test fails because it's trying to call the setter (it's read-only, same as is on `IWeb` and `IList`.

The unit test is effectively this:

```
// get first item and fetch the role assignments
var first = myList.Items.First();
await first.GetAsync(i => i.RoleAssignments);

Assert.IsTrue(first.RoleAssignments.Count() > 0);
```

And this is the error thrown:

```
    Test method PnP.Core.Test.SharePoint.ListItemTests.GetRoleAssignmentsTest threw exception: 
    System.ArgumentException: Property set method not found.
```

Implementation in `ListItem` is this:

```
public IRoleAssignmentCollection RoleAssignments { get => GetModelCollectionValue<IRoleAssignmentCollection>(); }
```

Have updated `RoleAssignment` with the property attribute to describe how to fetch an instance of itself:

```
[SharePointType("SP.RoleAssignment", Target = typeof(ListItem), Uri = "_api/Web/Lists(guid'{List.Id}')/Items({Parent.Id}/RoleAssignments/GetByPrincipal({PrincipalId})", Get = "_api/Web/Lists(guid'{List.Id}')/Items({Parent.Id}//RoleAssignments", LinqGet = "_api/Web/Lists(guid'{List.Id}')/Items({Parent.Id}/RoleAssignments")]
```

I believe, from looking at `JsonMappingHelper.cs::FromJsonRest` that it's because `IListItem` inherits `IExpandoDataModel` and this is preventing the code branch from going down to the point where it needs to add the `__deferred` URL to the datamodel so it's able to fetch and appropriately set the property, so it's trying to use reflection to call the property setter (which shouldn't exist anyway).

jansenbe:
@jimmywim , apologies for the late response, been quite hectic weeks. Would you have time for a quick call to run over things, it's probably easier to explain via a call than via writing, especially since we did a quite big internal refactoring last week and still need to update contributor docs

[...]

PR #343, opened 2021-03-04
List Item permissions / role assignments

jimmywim:
Re: #335 

Added ability to add/enumerate/remove list item permissions using role assignments.

Followed the pattern used in Lists, so the same techniques used for setting up permissions on lists should work on list items now.


jansenbe:
Thanks for sharing this @jimmywim !

Issue #298, opened 2021-01-20
Build a "lightweight" CSOM client that we can use for our required CSOM operations


jansenbe:
## Category

- [x] Feature request

## Describe the feature

Some operations do require CSOM, currently we've a solution that's based on coming up with the XML payload via string manipulation. This works but has limitations:

- does not support batching of requests
- hard to maintain

What if we'd had a minimal CSOM client implementation that could use information from our model (we can add additional model attributes for this purpose) and then can be used to load data or run methods via CSOM.

mgwojciech:
Thanks! That was fun :)

PR #345, opened 2021-03-07
CSOM Helper - update list item

mgwojciech:
Refactored ListItem to use CSOM helper for CSOM calls


jansenbe:
Thanks a ton @mgwojciech ! Will process this one asap

mgwojciech:
Oh and I think it can be associated with #298

PR #349, opened 2021-03-10
Provisioning Taxonomy field + Content type creation 

mgwojciech:
Hello,
Please note I updated few mock files (I run few tests in integration mode against my tenant). I also changed the unit test approach a little bit to isolate the PnPContext. I don't rely on existing tests so much as I try to reduce the dependency to PnPContext. Instead I'm testing for example serialization in isolation and CSOM Request by comparing actions/paths to expected values. Of course existing tests should pass.

Let me know what do You think about it.


jansenbe:
Thanks again @mgwojciech !

PR #375, opened 2021-03-29
Update listitems-intro.md

dgtheninja:
Added missing .AsRequested() extension method to get results of queries rather than all items


jansenbe:
Nice catch @dgtheninja , thanks for the PR to fix this.

PR #397, opened 2021-04-16
Update sites-intro.md

wonderplayer:
Web doesn't have `HubSiteId` property, but site does.


jansenbe:
@wonderplayer : good catch, thanks for the fix!

PR #404, opened 2021-04-21
Update IList.cs

wonderplayer:
Incorrect description of URL property. Guess it contains absolute URL?
Tried to get URL to really see what it has, but for some reason, I just couldn't get that property to load.


jansenbe:
Thanks @wonderplayer , that property indeed needs to be removed as it's not something returned by the API's. The rootfolder property is the one to use get the list url

PR #442, opened 2021-05-12
Allow to Set the GroupId we got from SharePoint in existing Context for Group

czullu:
My question in https://github.com/pnp/pnpcore/discussions/440 was if there is something against reusing the Context i have to work with Groups as it will be possible with this change.


jansenbe:
Hi @czullu , I've fixed this internally as you as a developer should not have to manually populate the metadata on the model objects. See https://github.com/pnp/pnpcore/commit/8442e95dcb3f3b3d2d39b5b513ddebbc5c984a2f for the commit with the fixes and test case. If you still experience issues then please open an issue. Thanks for making us aware of this bug!

Issue #428, opened 2021-05-06
IListItem DisplayName + UniqueId


DaleyKD:
## Category

- [x] Domain model extension

## Describe the domain model extension
Specifically, I need DisplayName and UniqueId, but really, I'd love to see `IListItem` fleshed out a bit more, like: https://github.com/pnp/pnpcore/blob/568a61199e11ba640d1fe91ade980225ea5c7be9/src/generated/SP/Internal/ListItem.cs

### Model
**Model**: IListItem
**Properties**: DisplayName, UniqueId

The interesting thing is that UniqueId is in the properties bag, but DisplayName isn't.  When looking at the returns of the REST endpoint, I can see why.

(The truth is, I'd love to see (almost NEED it) `IListItem` fleshed out using the majority of properties that MS.SP.Client.dll has, and that you have here: https://github.com/pnp/pnpcore/blob/568a61199e11ba640d1fe91ade980225ea5c7be9/src/generated/SP/Internal/ListItem.cs)

### APIs
API: `https://<tenant>.sharepoint.com/sites/<site>/_api/web/lists(guid'<listId>')/items(<listItemId>)?$select=UniqueId`
and
API: `https://<tenant>.sharepoint.com/sites/<site>/_api/web/lists(guid'<listId>')/items(<listItemId>)/displayname`

### Additional Information
[List Item SharePoint REST Documentation](https://docs.microsoft.com/en-us/previous-versions/office/developer/sharepoint-rest-reference/dn531433(v=office.15)?redirectedfrom=MSDN#listitem-properties)

Here's something SUPER crazy:
![image](https://user-images.githubusercontent.com/7206282/117315311-30d65600-ae4d-11eb-96b6-b428d31094e0.png)

The docs say that `DisplayName` is returned as `Title`, but if I request `Title`, it's almost always `null`.

jansenbe:
@DaleyKD : think this one can be closed now, if not then let me know and I reopen

jansenbe:
@DaleyKD : I took a peek at the source code and the docs are not accurate. What is returned depends:

- If a field with internal name "Title" exists and the value is not empty/null that's used
- Special handling for survey, discussion board and link items --> should not be relevant in your case
- Value of the "First field with TitleField="TRUE" in their manifest"

DaleyKD:
[...]

Issue #420, opened 2021-05-04
IListItem FileSystemObjectType and File / IsFile()


DaleyKD:
## Category

- [x] Domain model extension

## Describe the domain model extension
We frequently will load `IListItemCollection` and `.Include()` (sorry for old CSOM) the following:

``` csharp
    item => item.ContentType.Id,
    item => item.ContentType.Name,
    item => item.FileSystemObjectType,
    item => item.Folder.ServerRelativeUrl,
    item => item.File.CheckedOutByUser,
    item => item.File.Exists,
    item => item.File.Name,
    item => item.File.ServerRelativeUrl,
    item => item.File.TimeCreated,
    item => item.File.TimeLastModified,
    item => item.File.Title,
    item => item.File.Length
```

We then condition on the `FileSystemObjectType` to determine if it is a Folder or File (or neither).

### Model
**Model**: IListItem
**Properties**: File, FileSystemObjectType

### APIs
API: `https://<tenant>.sharepoint.com/sites/<site>/_api/web/lists(guid'<listId>')/items(<listItemId>)`
and
API: `https://<tenant>.sharepoint.com/sites/<site>/_api/web/lists(guid'<listId>')/items(<listItemId>)/file`

_Not perfect_
API: `https://graph.microsoft.com/v1.0/sites/<siteId>/lists/<listId>/items/<itemId>/driveItem`

### Additional Information
[List Item Microsoft Graph Documentation](https://docs.microsoft.com/en-us/graph/api/resources/listitem?view=graph-rest-1.0)

DaleyKD:
Currently having to make an extra call using: `await context.Web.GetFileByServerRelativeUrlAsync(fileUri);`  Since we get a bunch of list items in a single call, having to do this seems expensive (one additional call per item... that hits REST a lot).

DaleyKD:
Playing around with REST right now, it appears that the following works:
`https://<tenant>.sharepoint.com/sites/<site>/_api/web/lists(guid'<listId>')/items(<listItemId>)?$select=id,FileSystemObjectType`

Side note: not sure how this would work with using `LoadListDataAsStreamAsync`

jansenbe:
Implemented. See https://github.com/pnp/pnpcore/commit/f4da869d751af29602d47b988dc19a2222367cc7

DaleyKD:
@jansenbe - I appreciate the added functions, but this doesn't fix my actual need.

[...]

Issue #419, opened 2021-05-04
IListItem Content Type


DaleyKD:
## Category

- [x] Domain model extension

## Describe the domain model extension
Get list item's content type + content type ID

### Model
**Model**: IListItem
**Properties**: ContentTypeId, ContentType

### APIs
API: `https://graph.microsoft.com/v1.0/sites/<siteId>/lists/<listId>/items/<itemId>`
and
API: `https://graph.microsoft.com/v1.0/sites/<siteId>/lists/<listId>/items/<itemId>/contentType`

API: `https://<tenant>.sharepoint.com/sites/<site>/_api/web/lists(guid'<listId>')/items(<listItemId>)`
and
API: `https://<tenant>.sharepoint.com/sites/<site>/_api/web/lists(guid'<listId>')/items(<listItemId>)/contentType`

### Additional Information
[Content Types Microsoft Graph Documentation](https://docs.microsoft.com/en-us/graph/api/resources/contenttype?view=graph-rest-1.0)

I only specifically need `ID` and `Name` right now, but would like to load it when I load the `IListItemCollection`.

DaleyKD:
Right now, to get the name of the ContentType, I'm having to rely on the entire `Values` bag and use "ContentType".  I'd rather be able to load this explicitly.

jansenbe:
@DaleyKD : I've added support for the loading the content type when using LoadListDataAsStream. It only gets you a content type with name and id properties though. With the other changes you've done I think we now can get the content type populated for all listitem data retrieval approaches...You're good with closing this one or did I miss something?

DaleyKD:
@jansenbe - Initial tests are great.  Closing now.

Issue #413, opened 2021-04-28
IListItem.GetFolder return Parent IFolder instead of the current folder


seanix99:
## Category

- [x] Bug

## Steps to reproduce

// get a IListItem from what method you want for a folder 

// Here item represent: http://test.sharepoint.com/sites/test/documents/TestFolder

IListItem item = ....

If (item.IsFolder())
{
  IFolder folder = item.GetFolder(); 
  // here instead of receiving an IFolder object that represent the TestFolder, we receive an IFolder that represent the root folder.
}

## Expected behavior

GetFolder(), should return an IFolder representing the current ListItem when its a folder.

## Environment details (*development & target environment*)

> **DELETE THIS LINE BEFORE SUBMITTING** - Complete the following (*if ignored you'll be prompted for it before we can address your issue... save the time and provide it in your initial submission*).

- **SDK version**: 1.1.0
- **Framework**: .Net Core

DaleyKD:
For what it's worth, if I have any luck, I'm working on a PR that'll include IListItem properties, one of which is Folder.  It would possibly satisfy this issue.

I'll have more info after writing unit tests tomorrow.

DaleyKD:
@jansenbe & @seanix99 , what's the status with this?

If you grab a nightly, you can load the actual Folder object of the ListItem using the `IListItem.Folder` property.  And then you can use `.GetFolder()` to grab the folder to which the listItem belongs based on `FileDirRef`

Should we rename the function to be a bit less confusing?  `GetParentFolder`?  `GetContainerFolder`?  🤷‍♂️

jansenbe:
GetParentFolder makes sense to me...I'm doing the rename, so closing this issue. Thanks raising it again @DaleyKD

PR #452, opened 2021-05-18
Extending IListItem

DaleyKD:
* Added a bunch of properties to `IListItem`.
* Allow selectors for loading an `IListItem`.
* Added `GetDisplayName` functions.

Passes all unit tests, and greatly extends some of the functionality for `IListItem`s.

There are two things I'm not in love with here:
1. To load data using the DataStream, I parse the returned data twice.  The first time is to try and map the results to the class properties of an `IListItem`.  The second time is to fix the overflow dictionary with proper data types (like `IFieldUserValue`). This would have a minimal impact on performance, and I'm sure there's probably a better way to do it.
1. In order to work with the first pass thru for DataStream (where the JSON returned has the "Rows" element), I had to put a special use case handling.  DataStream returns **"ID": 153** for an `IListItem`, where everything else returns **"Id": 153**.  Being case-insensitive for ALL properties is bad, but this one instance needed to be case-insensitive.


DaleyKD:
Used to help fix #428 , though the DisplayName documentation is still wrong?
Fixes #420 , though there's no way to load Files via DataStream. 😞 
Fixes #419 , though not really when using DataStream.
Helps #413 .  Not sure the intent of `GetFolder`.  The `Folder` property is intended to get the Folder object if the `IListItem` is a folder.  Perhaps we rename `GetFolder` ?

DaleyKD:
@jansenbe  - Man, that's so freakin' cool.  Thanks for letting me be a contributor to this.  It feels kinda neat. :)

I'll grab the nightly tomorrow and plug it into our dev branch code to see what comes of it.

Do you know if there's a MAJOR performance difference between the CAML and LoadData (outside of SharePoint REST not giving us paging info in CAML)?  In my ideal world, I'd be able to pull the `IListItem.Folder` in a single request.

Issue #306, opened 2021-01-27
Introduce concept to allow additional headers being set on a graph request


jansenbe:
## Category

- [x] Feature request

## Describe the feature

Some graph API calls do require additional headers to be specified, we need a model that allows the additional headers to be specified at the model and/or provided in the API call

jansenbe:
Working on this one

PR #459, opened 2021-05-21
Enhance apirequest to allow setting http-header

czullu:
In order to do a Graph PATCH Request i needed the ability to pass in the IF-MATCH Header with the etag.
The enhancement allows this by adding a optional Dictionary to APIRequest and APICall which then will be respected in BuildMicrosoftGraphBatchRequestContent.

I did not change BuildSharePointRestBatchRequestContent as I could not see the benefit of doing so and 
the IF-MATCH=* in SP-REST has already the todo flag set and will be quite a large change if etag should be respected by default.


jansenbe:
Thanks @czullu . This also relates to https://github.com/pnp/pnpcore/issues/306

PR #461, opened 2021-05-21
Update listitems-intro.md

cesarhoeflich:
When performing some queries with CAML Query and keeping the 'p => p.Items' parameters, the complete list return has always occurred.


DaleyKD:
I see your point!

jansenbe:
@cesarhoeflich : good catch!

PR #462, opened 2021-05-21
Add the ability to load default properties for web/site

DaleyKD:
* Batch load the web and site properties on Init. (This requires me to redo all the mock data).
* Added the ability to load default properties when spinning up a new PnPContext via the factory.
* Fixed the FolderTest by removing a trailing slash.
* Fixed some Web/Site Tests (based on my new mock data, which is a fresh site).

@jansenbe - I'm truly sorry this PR is massive.  I hate the idea that I had to redo all of the mock data because I was batch loading the site/web at the beginning.  I understand if this PR is rejected.

A couple things:
* I **fully** expect the tests to fail.  There are three specific ones that I need you to personally run:
    * Possibly anything Syntex
    * This includes the PageTests for VivaTopics (I have no idea how to set this up) `TopicPageClone`
    * There are two in ListItemTests that have hardcoded Taxonomy IDs.  Obviously I couldn't do that.  (`SpecialFieldCsomTest` and `SpecialFieldRestUpdateTest`)
* We now have orphaned .response files.  Because of the new batch load, some of the last response files were unmodified, since they weren't used anymore.
* The clone only supports simple properties, like bool, string, Guid, DateTime, enum, etc.  I would LOVE to find a way to clone some collections, specifically `IWeb.AllProperties`.  It also wouldn't support complex objects (e.g. - `IBasePermissions`)


jansenbe:
Changed code files in this PR:

 src/sdk/PnP.Core.Test/Base/PnPContextTests.cs      |  423 ++-
 .../Services/Core/IPnPTestContextFactory.cs        |    8 +
 .../Services/Core/TestPnPContextFactory.cs         |   36 +-
 src/sdk/PnP.Core.Test/SharePoint/FoldersTests.cs   |    5 +-
 src/sdk/PnP.Core.Test/SharePoint/ListItemTests.cs  |    2 +
 src/sdk/PnP.Core.Test/SharePoint/SiteTests.cs      |    8 +-
 src/sdk/PnP.Core.Test/SharePoint/WebTests.cs       |    6 +-
 src/sdk/PnP.Core.Test/Utilities/TestCommon.cs      |   35 +-
 .../SharePoint/Core/Internal/RegionalSettings.cs   |    3 +
 .../SharePoint/Core/Public/IRegionalSettings.cs    |    5 +
 .../PnP.Core/Services/Core/PnPContextFactory.cs    |  141 +-
 .../Services/Core/PnPContextFactoryOptions.cs      |   15 +-

jansenbe:
Hey @DaleyKD ,  thanks for this PR. I've reviewed the code and I'm not going to merge this PR as is, although there's a ton of good work in here that I'll reuse. The two main reasons why this PR can't be merged are:

- In PnPContextFactory.InitializeContextAsync the separate interactive call for the Web was on purpose: when using the batch endpoint the urls are case sensitive, while doing an interactive request they're not. So we do one interactive request to grab the correct cased web url. This implies that we'll always need 2 requests...one interactive to ensure the url is correctly cased, one batched call for all the other initialization work. The fact that you merged both into one call meant you had to regenerate all mock files, which is not needed since we need 2 separate calls
- You're using the PnPContextFactoryOptions to specify the additional web/site properties, which is good, but since these settings can also be provided via a configuration file using a linq expression is not ideal

So overall great work, I've isolated the actual code changes you did and will reuse those when implementing this feature request (as it's a great idea to improve overall performance in apps that do operate on many sites). Will also ensure to credit you for the work in the changelog.

Thanks!

DaleyKD:
I totally understand not taking this PR.  It was huge and I knew it may have issues.

1. Ah, that's **_crazy_** that the batch endpoint is case sensitive.  (Maybe SharePoint REST should change that. :wink: )
2. I knew that the .config files part was going to be a "problem."  I was just like "-shrug-... you gotta do it in code."  Heh.  Maybe we can write some kind of converter to take a comma-delimited string and convert it to Expressions?  That's a decent amount of work, but I don't know what else to do.  (Ooooooor.... we could do it this way and support those two properties in config files at a later date.  Hehehehe.)

Thanks for doing this.  It irks me QUITE a bit that I have to make 3 calls (1 to Web, 1 to Site, then 1 to batch to load additional site and web props) EVERY time I spin up a context, just to get started doing actual work.  :smile:

DaleyKD:
Also, knowing this, can we add a little extra comment in the code explaining that the case-sensitivity is because of the batch endpoint?  I did see the comment regarding updating the `context.Uri`, but there was no explanation to the why of the non-batched call.

jansenbe:
Implemented via https://github.com/pnp/pnpcore/commit/794e520598986339f9cadead7a93e1c8c8467ec9, thanks for the inspiration on this one @DaleyKD !

DaleyKD:
NICE!  Added a quick comment re: `AllProperties`

Thank you.

PR #463, opened 2021-05-21
Can now load an IFile directly off of an IListItem using LoadAsync

DaleyKD:
Oooo, finally an easy change. :)

If my previous PR is accepted, we may need to rerun the tests on here so the .response files are updated.


jansenbe:
Thanks a ton @DaleyKD !

DaleyKD:
DOH!  I missed the one Mocking line commented out.  Sorry.

Issue #425, opened 2021-05-05
IContentType Fields


DaleyKD:
## Category

- [x] Domain model extension

## Describe the domain model extension
I appreciate that `IContentType` has `IFieldLinkCollection` on it. However, I'd do better with `IFieldCollection` on it, too.

### Model

**Model**: IContentType
**Properties**: `public IFieldCollection Fields { get; set; }`

### APIs
API: `https://<tenant>.sharepoint.com/sites/<site>/_api/web/lists(guid'<listId>')/contentTypes('<contentTypeId'>)/fields`

_I don't see a good equivalent in Graph_

jansenbe:
Thanks for the PR @DaleyKD

PR #464, opened 2021-05-22
Can now load Fields for a ContentType

DaleyKD:
Doesn't have a `[SharePointType]` attribute to fully flesh this out, but it does load when we want to load them as part of a `.QueryProperties()`.

Should satisfy #425 .


jansenbe:
Thanks @DaleyKD !

Issue #479, opened 2021-06-09
IListItem - Versions


DaleyKD:
https://s-kainet.github.io/sp-rest-explorer/#/entity/Collection(SP.ListItemVersion)
https://s-kainet.github.io/sp-rest-explorer/#/entity/SP.ListItemVersion

@jansenbe - Would you assign this to me?  I'll try to get to it this week.  Looks like this was missing and we'll need it for our product soon.

jansenbe:
Thanks for the great work @DaleyKD , just merged your PR so closing the issue

PR #481, opened 2021-06-10
Added ListItemVersions

DaleyKD:
As I was speaking with our team about our use cases, we realized there's a chance that there could be hundreds of ListItemVersions.

I tried to design ListItemVersions similarly to ListItems to allow for paging and filtering.

Two things I did not do:
- Finish the `DeleteObject` function (https://s-kainet.github.io/sp-rest-explorer/#/entity/SP.ListItemVersion/func/DeleteObject)
- Set up a unit test for an IFile/IListItem so that we could test the `IFileVersion` property

#479


DaleyKD:
Related to #479

PR #483, opened 2021-06-11
Fix Download of FileVersion Binaries

DaleyKD:
@jansenbe , this one got a bit hairy.

- For some reason, my unit test `GetListItemVersionFileVersionContentAsyncTest` works fine with real data, but not mock?
- To get this to work, I had to change `IFileVersion.Id` to `IFileVersion.ID`, which I'm NOT a fan of. I had to do this because of the way `TokenHandler` gets the `Parent.Id` out of the metadata bag. That is set in `JsonMappingHelper` by comparing the name of the `JsonProperty` to `SharePointKeyField.Name`:
``` c#
if (string.IsNullOrEmpty(idFieldValue) && property.Name.Equals(entity.SharePointKeyField.Name))
{
    idFieldValue = GetJsonPropertyValue(property).ToString();
}
```
I consider that a hair buggy, because it seems that when the `EntityInfo` is composed, it doesn't look at `KeyPropertyAttribute` and get the `KeyPropertyName`, to then hopefully look at the `SharePointPropertyAttribute`.
- I had to stop working on `IListItemVersion.FileVersion.GetContentAsync` because of the same reason. `IListItemVersion.Id` is REALLY `VersionId`, but the `JsonMappingHelper` doesn't know how to properly decipher that, so it never puts the VersionId in the `restId` property of the metadata bag.  This stops the `TokenHandler` from properly giving `Parent.Id` a value.

I hope that makes sense.  Simply put, if the SharePoint REST property (case sensitive) differs from `[KeyProperty(nameof(Id))]` (whatever is the `nameof()`), then we will get an UnresolvedTokens exception.

If we can find a way around that, then this will be 🏅 .


jansenbe:
Thanks a lot for this PR @DaleyKD , great to see all these good PR's!

About your questions:

- the test case that didn't work in offline mode: you've used the context id across different contexts...the test then works live, when fetching the offline data for context 2 (as an example) this fails because the code that uses context did persist offline with context id 2, effectively overwriting the previous offline responses with test errors consequently
- About Id versus ID: when I merged my work around list item/page comments I already made the field retrieval checks case insensitive, this also solved this issue
- About the issue with key field names when they're mapped to another name (SPO returned versionid while we used id): this is solved by checking on SharePointName instead of Name, the first contains the actual property name as returned by SPO REST, the latter contains the name of the property used in the model

DaleyKD:
- I can't believe I missed the context number.  I guess my eyes were going blurry by that point. :)
- Nice change on the case insensitivity!
- Thanks for making that change JsonMappingHelper.

PR #554, opened 2021-09-24
Added _CommentCount to BuiltInFields

avuorine:
_CommentCount is processed as a lookup field and ends up being null every time when using LoadListDataAsStreamAsync. Adding it to BuiltInFields fixes this.


jansenbe:
Thanks @avuorine , I'll process this soon.

Issue #556, opened 2021-09-29
Classic Publishing Page to Modern Transform - Page Analyser Error


pkbullock:
## Category

- [x] Bug

## Describe the bug

Page Transformation Analyser error when transforming from Classic Publishing Page to Modern Page.
The field its trying to read is a FieldUrlValue rather than string.

SharePointMappingProvider.cs

![image](https://user-images.githubusercontent.com/8781041/135338863-faaf97df-b8c3-4b8c-86bd-6f5fdab78be9.png)

Error:

  Message: 
    Test method PnP.Core.Transformation.SharePoint.Test.TransformationTests.SharepointTransformAsync threw exception: 
    System.InvalidCastException: Unable to cast object of type 'Microsoft.SharePoint.Client.FieldUrlValue' to type 'System.String'.
  Stack Trace: 
    ListItemExtensions.TryGetFieldValue[TFieldValue](ListItem item, String fieldName, TFieldValue& fieldValue) line 98
    SharePointMappingProvider.AnalyzePublishingPageAsync(PageTransformationContext context, File pageFile, ListItem pageItem, SourcePageInformation page) line 1586
    SharePointMappingProvider.AnalyzePageAsync(PageTransformationContext context, File pageFile, ListItem pageItem, SourcePageInformation page) line 1204
    SharePointMappingProvider.MapAsync(MappingProviderInput input, CancellationToken token) line 141
    DefaultPageTransformator.TransformAsync(PageTransformationTask task, CancellationToken token) line 82
    TransformationTests.SharepointTransformAsync() line 76
    ThreadOperations.ExecuteWithAbortSafety(Action action)

## Steps to reproduce

Transform Publishing Page Classic to Modern
Run Unit test SharePointTransformAsync

## Expected behavior

Magically works :-)

pkbullock:
So in the old tool, there was an extension method for this, could this be used to fix it, asking rather than "playing" as need to learn how new one works :-)

![image](https://user-images.githubusercontent.com/8781041/135339437-1cc55cfc-16b7-451b-8024-017f309df86a.png)

pkbullock:
This has been resolved.

PR #558, opened 2021-09-29
Minor fix for resolving the page layout file url

pkbullock:
The field type isn't a string thus blew up.

Fixed this and progresses.


pkbullock:
Ref #556

jansenbe:
Thanks for testing and creating this fix @pkbullock!

Issue #573, opened 2021-10-11
Transforming pages in a subsite referring to images in site collection fails


pkbullock:
## Category

- [x] Bug

## Describe the bug

Error:

Message: 
    Test method PnP.Core.Transformation.SharePoint.Test.TransformationTests.SharePointTransformAsync threw exception: 
    Microsoft.SharePoint.Client.ServerException: Server relative urls must start with SPWeb.ServerRelativeUrl
  Stack Trace: 
    ClientRequest.ProcessResponseStream(Stream responseStream)
    ClientRequest.ProcessResponse()
    ClientRequest.ExecuteQueryToServerAsync(ChunkStringBuilder sb)
    ClientRequest.ExecuteQueryAsync()
    ClientRuntimeContext.ExecuteQueryAsync()
    ClientContext.ExecuteQueryAsync()
    ClientContextExtensions.ExecuteQueryImplementation(ClientRuntimeContext clientContext, Int32 retryCount, String userAgent) line 130
    ClientContextExtensions.ExecuteQueryRetry(ClientRuntimeContext clientContext, Int32 retryCount, String userAgent) line 93
    SharePointFunctionsService.PersistImageFileContent(String sourceAssetRelativeUrl, ClientContext context, Int32 fileChunkSizeInMB) line 753
    SharePointFunctionsService.RetrieveImageFileRelativePath(String sourceAssetRelativeUrl, ClientContext context) line 727
    SharePointFunctionsService.ReturnCrossSiteRelativePath(String imageLink) line 702

## Steps to reproduce

Scenario:

- Multi-lingual Intranet so English/French. English is a sub site called "EN"
- Pages that contain references to files in <SiteCollection>/PublishingImages library

Paths:
https://capadevtest.sharepoint.com/sites/PnPSauce/en/Pages/Article-SPO-New-London-Flagship-Store.aspx
https://capadevtest.sharepoint.com/sites/PnPSauce/PublishingImages/allspice-aromatic-chilli-33615.jpg?ctag=211010

Run transformation and the "PersistImageFileContent" tries to retrieve the file from the current Web. The function needs to cater for the current site collection as well. 

Looking at the code, the caller RetrieveImageFileRelativePath prob needs to cater for this scenario.

Issue #561, opened 2021-10-03
Publishing Pages Target Page URI Transformer issue


pkbullock:
## Category

- [x] Bug

## Describe the bug

When transforming I get a rest error because the target page URL is not well formed for the target page. E.g.

Source URL:
https://capadevtest.sharepoint.com/sites/PnPSauce/en/Pages/Article-SPO-New-London-Flagship-Store.aspx 

Target URI: /sites/PnPKatchup/en/Pages/Migrated_Article-SPO-New-London-Flagship-Store.aspx

the "en" and "pages" library fragments of the URL shouldn't be there.

![image](https://user-images.githubusercontent.com/8781041/135769587-e288a089-6b2e-44f1-9e1c-63fe48df3ed4.png)

Error Message:

 Message: 
    Test method PnP.Core.Transformation.SharePoint.Test.TransformationTests.SharepointTransformAsync threw exception: 
    PnP.Core.SharePointRestServiceException: SharePoint Rest service exception
  Stack Trace: 
    BatchClient.ProcessSharePointRestBatchResponseContent(Batch batch, String batchResponse) line 1220
    BatchClient.ProcessSharePointRestBatchResponse(SPORestBatch restBatch, String batchResponse) line 1162
    BatchClient.ExecuteSharePointRestBatchAsync(Batch batch) line 1013
    BatchClient.ExecuteBatch(Batch batch) line 339
    BaseDataModel`1.BaseRetrieveAsync(ApiCall apiOverride, Func`2 fromJsonCasting, Action`1 postMappingJson, Expression`1[] expressions) line 556
    Web.GetFileByServerRelativeUrlAsync(String serverRelativeUrl, Expression`1[] expressions) line 513
    DefaultPageGenerator.GenerateAsync(PageTransformationContext context, MappingProviderOutput mappingOutput, Uri targetPageUri, CancellationToken token) line 137
    DefaultPageTransformator.TransformAsync(PageTransformationTask task, CancellationToken token) line 86
    TransformationTests.SharepointTransformAsync() line 76
    ThreadOperations.ExecuteWithAbortSafety(Action action)

pkbullock:
Note this issue occurs after PR #563

pkbullock:
Isolated the fault to an issue in the ResolveAync method in SharePointTargetPageUriResolver class.

jansenbe:
Closing for now, will re-open if this effort will be picked up again

PR #574, opened 2021-10-11
Couple of fixes

pkbullock:
- Ref #573 Fixed issue with images referring to parent web 
- Ref #561 Fix target page location calculation if source originated on a sub web

Small steps forward for tonight. After these fixes I was able to transform page :-) 
This currently is online to online transformation.


jansenbe:
Thanks for merging this one @pkbullock , I've planned to work on modernization but so far did not get the time to do so

PR #606, opened 2021-10-30
Added scripts for easier coverage reports.

s-KaiNet:
As discussed at [Code coverage report running locally](https://github.com/pnp/pnpcore/discussions/603) - this PR adds necessary scripts to run code coverage reports from Powershell.    

I managed to make it cross-platform (I was mistaken that `reportgenerator` is windows only, it works fine on UNIX as well). Also added a small fix to tests so that they run on Unix, because Unix doesn't like `\\` in paths.


jansenbe:
Thanks again @s-KaiNet 🥇

PR #633, opened 2021-11-16
Fixing problem with "new Guid()" inside LINQ expressions.

### LINKED ISSUE #632: System.NotSupportedException when trying to LINQ by Guid id
## Category

- [x] Bug

## Describe the bug
Below code

```csharp
var list = context.Web.Lists.GetByServerRelativeUrl($"{context.Uri.LocalPath}/Shared Documents");
var view = list.Views.FirstOrDefault(v => v.Id == new Guid("<guid>"));
```
throws 
> System.NotSupportedException: Constant new Guid(value(...) is invalid. Only System.Linq.Expressions.ConstantExpression and System.Linq.Expressions.MemberExpression are supported

From the other side below code works just well:

```csharp
var list = context.Web.Lists.GetByServerRelativeUrl($"{context.Uri.LocalPath}/Shared Documents");
var id = new Guid("<guid>");
var view = list.Views.FirstOrDefault(v => v.Id == id);
```
This is a bit confusing because it's almost identical code.  

The problem is [here](https://github.com/pnp/pnpcore/blob/621061fa6cfe0dc0fdf5f11358a6375f3e2e1bab/src/sdk/PnP.Core/QueryModel/Query/ExpressionExtensions.cs?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L27), we don't handle [`NewExpression`](https://docs.microsoft.com/en-us/dotnet/api/system.linq.expressions.newexpression?view=net-5.0) scenario. It doesn't make sense to handle all "new" scenarios, but we can easily handle "new Guid" since it might be very common to get object by `Id == new Guid(string)` LINQ query.   

I will make a PR with a possible fix.

s-KaiNet:
This PR fixes #632

The idea is to compile lambda expression with `new Guid()` and invoke compiled delegate to get a real Guid instance: 

```csharp
Expression.Lambda<Func<Guid>>(expression).Compile().Invoke();
```

I searched for any info about `Expression.Compile()` performance and found [this SO answer](https://stackoverflow.com/a/62679981/434967), which indicates that performance is very good (600-700ms for 2M of iterations). I checked results on my own environment for .NET 5 and received similar results. Invoking a compiled delegate is also fast.


jansenbe:
Nice improvement @s-KaiNet 💪Also like the fact you've spend time on thinking about perf.

PR #646, opened 2021-11-19
Fixed issues with boolean conditions

### LINKED ISSUE #582: Where statement in Linq query is not taken in account when using bool conditions without specifying equal to true or false
In below snippets the where statements are ignored and that's because `list.HasUniqueRoleAssignments` is not correctly evaluated, writing `list.HasUniqueRoleAssignments == true` does work as expected.

```
var lists = await context2.Web.Lists.
    Where(list => list.HasUniqueRoleAssignments)
    .QueryProperties(list => list.HasUniqueRoleAssignments, list => list.Title).ToListAsync();
``` 

```
await foreach (var list in context2.Web.Lists.Where(list => list.HasUniqueRoleAssignments).AsAsyncEnumerable())
{
    await foreach (var item in list.Items.Where(item => item.HasUniqueRoleAssignments).AsAsyncEnumerable())
    {
        context2.Logger.LogInformation($"{list.Title} - {item.Id}");
    }
}
```

s-KaiNet:
While fixing the related issue I found a few more cases, when the behavior is unexpected/inconvenient.   

For example, all below `where` statements produced wrong REST filters (it applies to `FirstOrDefault` with filter expression as well):

```sql
where w.Title.StartsWith("Test") && w.Description == "Test"
where !w.Title.StartsWith("Test")
where w.Title.StartsWith("Test") -- < -- this one works though
where w.Description == "Test" && w.Title.Contains("Test") 
where w.HasUniqueRoleAssignments
etc...
```

To make it work you should always write them in a very strict way:

```sql
where w.Title.StartsWith("Test") == true && w.Description == "Test"
where w.Title.StartsWith("Test") == false
where w.Description == "Test" && w.Title.Contains("Test") == true
where w.HasUniqueRoleAssignments == true
```
Which looks kinda weird. 

The [DataModelQueryTranslator](https://github.com/pnp/pnpcore/blob/dev/src/sdk/PnP.Core/QueryModel/Query/DataModelQueryTranslator.cs) expects the above strict format and it was quite difficult to fix it without affecting a good part of the class. I fixed it a bit differently.

Instead, before generating an actual REST query, the initial LINQ Expression is normalized, i.e. all bool expressions are transformed to a binary format (`equals` _true_ or _false_).   

Having below initial filter:

> w.Title.Contains("Test") && w.Description == "Test" && !w.Hidden

the normalized version will be:
> w.Title.Contains("Test") == true && w.Description == "Test" && w.Hidden == false

and the final REST query:

> $filter=((startswith(Title,'Test') eq true and Description eq 'Test') and Hidden eq false)

You can see more samples inside [ExpressionNormalizerTest](https://github.com/pnp/pnpcore/compare/dev...s-KaiNet:fix-582?expand=1#diff-32904445ff016caf410e08b008bc307c494ac97405ae125237016e70ed1eb61c). 

With this PR you can now write LINQ filters with the boolean conditions without thinking about the right format.

Closes #582


jansenbe:
@s-KaiNet : @PaoloPia will be handling this PR and since we've been prepping and delivering a conference there's some delay, but rest assured, we'll process your PR as it's really good addition to PnP Core SDK.

PaoloPia:
Yes, I'm on it right now! Thanks!

PaoloPia:
And it is now merged! Thanks! Great job!

PR #653, opened 2021-11-23
Added taxonomy field related properties.

### LINKED ISSUE #596: Improve support for working with fields via a "typed" field collection
**Proposal for future improvement:**

Today web/list fields can be read and returned as a collection of `IField`. This works fine for basic fields but for more complex fields (eg taxonomy fields) part of the field definition (e.g. term set id) lives inside the field's `SchemaXml` property. What if there was a method `AsTypedFields()` on the current `IFieldsCollection` that would return the fields a collection of `ITextField`, `IBoolField`, `ITaxonomyField` ...all inheriting from a common `ITypedField`. This way developers don't have to parse the field xml themselves, they can just use a typed experience to work with the field properties.

These typed fields should also have the respective `Update` and `Delete` methods to allow for easy updating of the field.

s-KaiNet:
Closes #596 

I found out that we don't need to parse XML for the taxonomy field type. All taxonomy-related properties are now returned back as normal properties for a taxonomy field type. Thus as part of this PR, I simply added all those properties to the `IField` (added a new `#region` as for other field types in the `IField`).


jansenbe:
Thanks a ton @s-KaiNet , I'll process this one still today.

PR #668, opened 2021-12-01
Update listitems-intro.md

KoenZomers:
Clarifying that the option to use when trying to restrict the fields returned on list items to only specific custom fields should be option D, using CAML


jansenbe:
@KoenZomers : I'll merge this one but will slightly adjust the wording as it's about restricting the fields to load

Issue #300, opened 2021-01-21
Document the similarities/differences between CSOM and PnP Core SDK


jansenbe:
## Category

- [x] Feature request

## Describe the feature

PnP Core SDK feels a bit like CSOM but works not always the same...clarify the differences in documentation to assist folks moving over from CSOM.

jansenbe:
Waiting for the refactoring @PaoloPia is working on as this will further align the behavior between Core SDK and CSOM

olemp:
@jansenbe Is it possible now or will it be possible at any time to get the underlying client context (for CSOM)? I'm looking for the GetChanges API on the lists, and it seems to be "missing" so far.

s-KaiNet:
@jansenbe do you have any samples or plan how these docs should be structured? Otherwise, I can try to create the initial page with the similarities/differences between CSOM and PnP Core SDK. I don't think I will cover all possible scenarios, but maybe at least it will be a starting point. Better to have something to start with, than nothing :)

jansenbe:
Hi @s-KaiNet , I don't have any specific structure in mind...you do know both models very well, so I've no doubt your "starting point" already will be very good. I'll assign this one to you then. Thanks!!

jansenbe:
Thx @s-KaiNet 🚀🥇

PR #669, opened 2021-12-01
Added "csom vs pnp" docs.

s-KaiNet:
This PR adds a new docs page describing some similarities/differences between CSOM and PnP Core SDK. I'm not sure whether the related #300 issue should be closed with the PR, meaning is it full enough to close the issue. You decide ;). Anyway, we have something to start with. We can update this page with other samples in the future, if we see some misunderstanding among developers.

Also fixed a few minor things as part of this PR:
- simplified code in batching docs sample - `var batch = context.NewBatch();` 
- more semantically correct parameter name for the [IDataModelCollectionLoad](https://github.com/pnp/pnpcore/compare/dev...s-KaiNet:docs-300?expand=1#diff-f424725132fb94de888c5c6e552eda4dd9ca01023e3a423085b3985aef77f4f3)
- missing using statement for `ITermSet.cs` (this fix removed compiler warning)

### DIFF
--- a/docs/using-the-sdk/basics-batching.md
+++ b/docs/using-the-sdk/basics-batching.md
@@ -95,9 +95,9 @@ While above example showed how to add items in a batch it's also possible to bat
 
 ```csharp
 // Create a single batch that loads SharePoint Site and Web data + Teams Members. As this involves REST and Graph the net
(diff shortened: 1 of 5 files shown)


jansenbe:
Again a high quality contribution, thanks a ton @s-KaiNet !!!

PR #681, opened 2021-12-08
Update configuring authentication.md

JackStrap:
New way to register an application in Azure Active Directory with SharePoint PnP PowerShell Online cmdlets version 3.29.2101.0 and higher.


jansenbe:
Thanks for pointing that out @JackStrap , that's indeed a cleaner syntax.

Issue #722, opened 2022-01-14
Batching APIRequest seems broken


czullu:
## Category

- [x] Bug

## Describe the bug
While the Request for test1 still returns a result, the request for test2 returns null and shows a type initialization error.
However - when check loadPlansBatch.Requests[0].ResponseJson i can see the loaded Json-Data, so the call was fine and we got Data successfully.

```
using (var context = await pnpContextFactory.CreateAsync(new Uri("https://xyz.sharepoint.com/sites/TestGroupSiteWithPlanner")))
{
                 await context.Site.LoadAsync(s => s.GroupId);
                    Guid groupId = context.Site.GroupId;

                    var test1=context.Group.ExecuteRequest(new ApiRequest(ApiRequestType.Graph, $"groups/{groupId}/planner/plans"));

                    var loadPlansBatch = context.NewBatch();
                    var plannerResponse = context.Group.ExecuteRequestBatch(loadPlansBatch, new ApiRequest(ApiRequestType.Graph, $"groups/{groupId}/planner/plans"));
                    var Errors = await context.ExecuteAsync(loadPlansBatch, false);
                    if(!Errors.Any()&& plannerResponse.IsAvailable && plannerResponse.Result.Value!=null)
                    {
                        string test2 = plannerResponse.Result.Value;
                    }
}
```

## Steps to reproduce
Use a User Context against GroupSite with Planner and the call should return the assoicated Plans. While it does for single call, it does not for batch even when Data is actually visible in Batch-Object.

## Expected behavior
in earlier Version loading this data by Batch did work without an issue - would be nice it would do it again.

## Environment details (*development & target environment*)
DEV Version from today
Commit: a9320ab0ea5fab1b9103a84b9c86765ddfee6e9a [a9320ab]
Parents: cfb46c7de6, 41b2475123
Author: Bert Jansen <jansenbe@users.noreply.github.com>
Date: Friday, 14 January 2022 09:27:38

czullu:
@jansenbe That's how the object looks in Debugger 
![image](https://user-images.githubusercontent.com/31336529/149599895-b1bb17d4-f893-46bb-82bc-06a6f18fdc6a.png)
while i can see that ExecuteRequestBatchAsync ->RawResultsHandler get called and get's the json response i somewhat fail to see when BatchSingleResult -> result should be set. in Breakpoint BatchSingleResult.Result in object BatchRequest i still see the return value but in code we would return result..

jansenbe:
[...]

PR #723, opened 2022-01-15
fix single batch not returning data

czullu:
@jansenbe  fix for #722 in order to return data for single requests


jansenbe:
@czullu : thanks for the work here, as mentioned in #722 I've fixed this issue in a different manner (aligned with an earlier similar fix done for the REST based flow). Thanks for reporting the issue and helping with fixing it!

Issue #727, opened 2022-01-20
Error Parsing Page because "addedFromPersistedData":"True" is expected to be boolean


czullu:
## Category

- [x] Bug

## Describe the bug
We see an error spreading on tenants which breaks loading pages in PageWebPart.FromHtml because deserialization of WebPartControlData fails because of attribute addedFromPersistedData beeing a string and not a boolean.

element.GetAttribute(ControlDataAttribute) returns
{"addedFromPersistedData":"True",....
instead of
{"addedFromPersistedData":true,

## Steps to reproduce
as mentiones above we recognized this issue currently on 4 tenants.

a very uggly test in  PageWebPart.FromHtml:
```
            string dataJson = element.GetAttribute(ControlDataAttribute);
            try
            {
                dataJson = dataJson.Replace("\"addedFromPersistedData\":\"True\",", "\"addedFromPersistedData\":true,");
                SpControlData = JsonSerializer.Deserialize<WebPartControlData>(dataJson, PnPConstants.JsonSerializer_IgnoreNullValues);
                controlType = SpControlData.ControlType;
                RichTextEditorInstanceId = SpControlData.RteInstanceId;
            }
            catch (Exception ex)
            {
                string err = ex.Message;
            } 
```
## Expected behavior
make JsonSerializer.Deserialize<WebPartControlData> or the class WebPartControlData accept both ways for addedFromPersistedData in order to Load Pages.

## Environment details (*development & target environment*)
current dev branch of today

czullu:
@jansenbe the best idea i can come up with currently to fix this would be to have setter as string and getter as boolean for addedFromPersistedData in Model WebPartControlData - if ok, can do a fix

jansenbe:
Thanks for logging this issue and providing the fix @czullu 🥇🚀

jansenbe:
Closing as the fix has been merged.

PR #728, opened 2022-01-20
introduce BoolJsonConverter in order to handle both appearances of addedFromPersistedData

czullu:
#727 created JsonConverter for the attribute in order to handle deserialization of string and bool.


jansenbe:
Thanks for this fix @czullu !

Issue #729, opened 2022-01-24
Load SiteUsers fails with Error on AadObjectId


czullu:
## Category

- [x] Bug

## Describe the bug
the line: await foreach (var user in pnpCoreSiteContext.Web.SiteUsers.AsAsyncEnumerable())
will fail with error "Property AadObjectId was not yet loaded" on MoveNext as soon as we hit a user not from AAD i.e. a system user.

   at PnP.Core.Model.TransientObject.GetValue[T](String propertyName)
   at PnP.Core.Model.Security.SharePointUser.get_AadObjectId()

The Error remains the same if i do a more explicit request
await foreach (var user in pnpCoreSiteContext.Web.SiteUsers.QueryProperties(u => u.Id, u => u.LoginName, u => u.PrincipalType, u => u.AadObjectId).AsAsyncEnumerable())

## Expected behavior
in absence of the attribute in the request response it should return null for AadObjectId.  

## Environment details (*development & target environment*)
dev branch

i will provide a fix for PnP.Core.Model.TransientObject.GetValue[T](String propertyName) by adding param bool allowedToBeAbsent=false and change that on true for get_AadObjectId

jansenbe:
@czullu : I'll have a look later today / tomorrow. Thanks for bringing this up.

czullu:
Hi Bert, do you Run apponly and get system\User in the results?
I can provide full example later today.

Holen Sie sich Outlook für Android<https://aka.ms/AAb9ysg>
________________________________

jansenbe:
@czullu : I did not try via app-only yet, something I'll give a try later today as well

czullu:
failing users are "c:0(.s|true", "c:0-.f|rolemanager|spo-grid-all-users/....", "i:0#.w|nt service\spsearch", "i:0i.t|00000003-0000-0ff1-ce00-000000000000|app@sharepoint",  "SHAREPOINT\system". 

i'm getting them app-only from a GROUP#0 site on my MsDev-Tenant with options.PnPContext.GraphFirst = true;
rest of options are default, changing options.PnPContext.GraphFirst = false does not change behavior as expected.

with code 
```
await foreach (var user in context.Web.SiteUsers.QueryProperties(u => u.Id, u => u.LoginName, u => u.PrincipalType, u => u.AadObjectId).AsAsyncEnumerable())
{
    string aadObj = user.AadObjectId != null ? user.AadObjectId : "";
[...]

PR #730, opened 2022-01-24
suggested fix for Load SiteUsers fails with Error on AadObjectId

czullu:
fix for Error #729 - allows property to be absent in response and return default(T) instead.


jansenbe:
Pending feedback on discussion #729

jansenbe:
@czullu : see my answer to the issue. This is not an issue in PnP Core SDK, so closing this PR.

Issue #809, opened 2022-04-03
Exception when loading Folder from ListItem


FelgeIO:
## Category

- [x] Bug

## Describe the bug

An exception is thrown when properties of a Folder - that is represented by a ListItem - are loaded and/or accessed:
> Specify the GraphGet/GraphGetLinq field of the ClassMapping property

## Steps to reproduce

The following method can be used to reproduce the issue (provided that at least one ListItem representing a Folder is contained within the list):

```
public async Task GetFolderChanges(IList list)
{
    var queryOptions = new ChangeQueryOptions() { Add = true, Update = true, Item = true };
    IList<IChange> listChanges = list.GetChanges(queryOptions);

    foreach (var listChange in listChanges)
    {
        if (listChange is IChangeItem changeItem)
        {
	    var itemId = changeItem.ItemId;
            var listItem = list.Items.GetById(itemId);

            if (listItem.IsFolder())
            {
                // Option 1:
                try
                {
                    var folder = await listItem.Folder.GetAsync(f => f.ServerRelativeUrl, f => f.Name, f => f.TimeLastModified);
                    // do something with folder
                }
                catch (Exception ex)
                {
                    // Exception: "Specify the GraphGet/GraphGetLinq field of the ClassMapping property"
                }

                // Option 2:
                try
                {
                    var folder = listItem.Folder;
                    await folder.LoadAsync(f => f.ServerRelativeUrl, f => f.Name, f => f.TimeLastModified);
                    // do something with folder
                }
                catch (Exception ex)
                {
                    // Exception: "Specify the GraphGet/GraphGetLinq field of the ClassMapping property"
                }
            }
        }
    }
}
```
 
Presumably, one could also just iterate over a list to find a Folder in order to reproduce the issue .. I simply stumbled upon this while writing some code to process changes of a list.

## Expected behavior

When the above code is run for a ListItems representing Files (listItem.IsFile() + listItem.File), everything works as expected: The requested properties (ServerRelativeUrl, Name, TimeLastModified) are loaded correctly and can be accessed. I expect that this would also be the correct behavior for ListItems representing Folders .. but instead, an exception is thrown.

## Environment details

- **SDK version**: 1.6.0
- **OS**: Windows 10
- **SDK used in**: Azure Function App (dotnet-isolated)
- **Framework**: .NET Framework 6.0
- **Tooling**: Visual Studio 2022

### Additional context

The logs show that the exception is thrown before the request is even sent to SharePoint.

[...]

PR #810, opened 2022-04-03
Added option to load Folder from ListItem

FelgeIO:
Issue ref: #809 

- Fixed exception when Folder properties are accessed (and/or loaded) through a related ListItem object


jansenbe:
Thanks for this fix @FelgeIO , really appreciated!

Issue #414, opened 2021-04-29
Implement cross site collection copy/move


jansenbe:
Current implementation supports copy/move inside the site collection, cross site collection requires using different APIs. See https://github.com/pnp/powershell/blob/dev/src/Commands/Utilities/CopyMover.cs for a sample implementation.

DaleyKD:
Is this a duplicate of #301  ?

jansenbe:
Good point, I'll close #301....one issue less :-)

MathijsVerbeeck:
Hi @jansenbe 

I'd like to assist on this issue.

jansenbe:
All yours @MathijsVerbeeck 🚀

MathijsVerbeeck:
@jansenbe I wanted to look into this, but I noticed that this already has been implemented, however not using the CreateCopyJobs API but using the _api/SP.MoveCopyUtil.CopyFileByPath API call. Do I still have to implement this using the CreateCopyJobs API?

jansenbe:
@MathijsVerbeeck : having a look at things, will reply shortly

jansenbe:
@MathijsVerbeeck : the existing implementation works indeed cross site but is limited to copying a single file (no list items) whereas the CreateCopyJobs API can do files/folders/listitems and can do things in bulk (hence they work async). I would propose to add the needed extension methods at `ISite` to support this feature:

- `CreateCopyJobs` and `CreateCopyJobsAsync`
- `GetCopyJobProgress` and `GetCopyJobProgressAsync`

Would that work for you?

MathijsVerbeeck:
@jansenbe : Seems good to me.

jansenbe:
Thanks once more of this contribution @MathijsVerbeeck ! Just merged it, so closing this issue.

PR #814, opened 2022-04-05
Added the Copy / Move to functionality on the ISite interface

MathijsVerbeeck:
Added the possiblity on the Site object to migrate files / folders in the current site collection or between different site collections. Also added the possibility to check the progress and a function that will wait until all the jobs have finished

Added functions:
- CreateCopyJobs(Async)
- GetCopyJobProgress(Async)
- EnsureCopyJobHasFinished(Async)

Issue ref: #414


jansenbe:
Wow, that was fast @MathijsVerbeeck! Thanks again for the great work, this is another great addition for PnP Core SDK!

MathijsVerbeeck:
@jansenbe I just thought about maybe adding some documentation/checks about the limitations (max 2gb/item, max 100 files and max 100gb in total) but al struggling about 'how' to check for the filesizes because that would mean that I'll have to retrieve every file and get the filesize of this file. This would generate quite a bit of traffic. Any suggestions on how to do this?

jansenbe:
@MathijsVerbeeck : I would just do this via documentation. SDK users that want to enforce this can perform size checks in their code, enforcing this on every SDK users will be overkill for most use cases I think. 

Just merged your PR, did a few tweaks but overall very good job done again! One thing that can trick contributors is the fact that the test project can access the internal classes from the SDK, so that's why I moved from `ICopyMigrationOptions` and an internal `CopyMigrationOptions` class to a public `CopyMigrationOptions` class. Other small change I did is ensuring the delay inbetween status checks is always configurable. Needed that to avoid having to wait during test cases.

PR #818, opened 2022-04-09
Added findFile functionality to Web, List, Folder

dennisvanniel:
Added functionality I needed for finding files in Web, List, or Folder, with optional use of wildcard (*) for matching partial filenames. It's modelled after the Find-PnpFile cmdlet in PnP PowerShell.

Haven't posted a message in the issue list first as I should have done (as I see  now in the contributing guidelines)
No documentation written, though I shall write if you choose to accept these extensions.


jansenbe:
Nice work @dennisvanniel ! Looks like a good addition, I'll take a closer look as of the 19th as I'm OOF currently. Maybe I can process this before that date, but just wanted to set expectations.

jansenbe:
Hi @dennisvanniel , I've today merged your PR, nice job 💪🚀. There's one change I did and that's removing the find files option in IWeb as that possible extremely slow as it runs over each folder (also the ones without a list associated). Users can still trigger this behavior via the IWeb's RootFolder, but it's best to that a user only uses this feature on a given folder or list.

Would be good if you could add some basic documentation for this, if not possible then let me know so I can create an issue to track this for the future. Thanks :-)

dennisvanniel:
Hi @jansenbe, 
I'll write some basic documentation. Should be finished by the end of next week.

PR #829, opened 2022-04-26
Allow custom Accept header with ApiRequest

KoenZomers:
Adding ability to allow for providing a custom accept header with odata=nometadata


jansenbe:
Good addition @KoenZomers . Since we came from odata=verbose for all requests we could not default custom requests to nometadata like we did for all the internal calls as that would break existing users. With this change existing users do have an option to choose for nometa.

PR #845, opened 2022-05-10
Changed GraphGroup constructor from Internal to Public

danielpastoor:
Error when trying to get group data from a team site:
No parameterless constructor defined for type 'PnP.Core.Model.Security.GraphGroup'

Fix:
Change the constructor of GraphGroup from Internal to Public.

And added the documentation for the fixed part.


jansenbe:
Nice catch on the constructor @danielpastoor , it was accidently left internal.

Issue #310, opened 2021-01-29
Add Reply To to Chat Message


pkbullock:
## Category

- [x] Feature request

## Describe the feature

- Add Reply To a chat message capability
- Add tests to validate the feature
- Add documentation with sample usage code

Reference: https://docs.microsoft.com/en-us/graph/api/channel-post-messagereply?view=graph-rest-beta&tabs=http#example-1-create-a-new-reply-to-a-chatmessage

## Contributing
We encourage and welcome contributions for the PnP Core SDK, if you are interested, please comment below to engage with the team. If you want to learn about how to get started, please refer to this guide: https://pnp.github.io/pnpcore/contributing/readme.html

The team is constantly improving the guidance on supporting contributors, if you have any feedback or need help please feel free to start a discussion.

MathijsVerbeeck:
@jansenbe  I could look into this.

jansenbe:
All yours :-)

jansenbe:
The great PR's keep on coming, thanks again @MathijsVerbeeck for providing these features!!!

PR #853, opened 2022-05-17
Added functionalities to list replies of a teams chat message and add new replies

MathijsVerbeeck:
Issue ref: #310 

@jansenbe I'll try my best and write documentation for this feature + the other latest features regarding Team Events + Team Chat mentioning later this week.


MathijsVerbeeck:
@jansenbe I've added batching functionalities.

I noticed that sometimes, when having a channel with in my case +- 170 messages, the message that I've just added will not be retrieved when just "running" the test. To make the tests work, I have to add a breakpoint on the getting of the messages and continue and the newly created message will be found. If you can't reproduce this issue or still have questions due to my explanation being kinda "vague", don't hesitate to contact me 😄

PR #859, opened 2022-05-27
Added some new Features

danielpastoor:
I added some new features.

Features:
1.
Check if a site exists by alias and full url
2.
Added option to create a team site when you use "Application" permissions.
3.
Changed TeamChannelTabConstants.cs location from internal to public so people can use it in there own projects.


jansenbe:
Thanks for the PR @danielpastoor . I've merged this one, but did do quite some changes, maybe you want to have a look at https://github.com/pnp/pnpcore/commit/0deb3db8314819b808cb50141fd6237845bf7ed4 to understand what I've changed. Key changes are: 
- some simplification: a method does one thing, creating a team site via creating a group is fine, in that same method also trying to create a Team is too much
- Removed some overloads
- Added missing sync method
- Moved around some the SiteExists methods
- From a coding style we always wrap a block in braces, so the `if (condition) then do x`  has been updated to `if (condition) then { do x }`

But please don't get me wrong here, super happy to these contributions come in as they add value for folks using the SDK. Taking in account above recommendations will help speed up merging of future PRs :-)

danielpastoor:
@jansenbe Thank you for the feedback. I see this as a learning curf for me. And I will improve the my coding for the next time. Thanks

Issue #221, opened 2020-11-02
Create model for supporting document sets


pkbullock:
## Category

- [x] Feature request

## Describe the feature

- Add the ability to call the APIs for working with document sets
- Unit Tests for calling the APIs
- If there are special requirements or permissions with this model, add to the documentation

## Contributing

We encourage and welcome contributions for the PnP Core SDK, if you are interested, please comment below to engage with the team. If you want to learn about how to get started, please refer to this guide: [https://pnp.github.io/pnpcore/articles/contributor/readme.html](https://pnp.github.io/pnpcore/articles/contributor/readme.html)

The team is constantly improving the guidance on supporting contributors, if you have any feedback or need help please feel free to start a discussion.

MathijsVerbeeck:
@jansenbe I could have a look into this, but creating / updating the properties within a document set such as for example adding default documents, creating/updating the welcome page, ... have always been acting a bit strange for me when doing this with code. Haven't tried this yet using the SharePoint REST Api though.

jansenbe:
Feel free to give it a go!

jansenbe:
Support for document sets was added by @MathijsVerbeeck, thanks a ton for the help 🥇

PR #866, opened 2022-05-31
Create model for supporting document sets (Add and update)

MathijsVerbeeck:
Issue ref: #221 

The deletion of the content-type is done by deleting the parent content type.

@jansenbe Thanks for assisting me / looking with me for the appropriate Graph API calls 😄 Also, in case of any questions, don't hesitate to DM me 😄


jansenbe:
Awesome work again @MathijsVerbeeck !!

PR #935, opened 2022-08-08
samples were referencing the wrong variable

HiltonGiesenow:
per the change, the samples were referencing an non-existing variable name


jansenbe:
@HiltonGiesenow : nice catch, thanks for the PR 💪

PR #937, opened 2022-08-11
Search sort direction

lroos:
Fixes a bug introduced by the switch from GET to POST search query by aligning the SortDirection enum with System.ComponentModel.ListSortDirection (Ascending = 0). This was previously passed as a string but POST model requires it as an int.


jansenbe:
Thanks for the fix @lroos 🚀👍

PR #978, opened 2022-09-28
Adding ability to specify ResourceBehaviorOptions when creating a Group

KoenZomers:
Added option to specify the [ResourceBehaviorOptions](https://learn.microsoft.com/en-us/graph/group-set-options#configure-groups) when creating a Microsoft 365 Group using the code like provided in this sample: https://pnp.github.io/pnpcore/using-the-sdk/admin-sharepoint-sites.html#basic-site-collection-creation-flow


jansenbe:
Thanks @KoenZomers, good addition. Given we call SPO Rest APIs for site creation when using delegated permissions this currently only works for application permissions. I've clarified this in the code comments and will also do in our docs.

PR #1010, opened 2022-10-28
Update pages-webparts.md

soebje:
Missing "page." before  "GetImageWebPartAsync" on line: 191


soebje:
Missing "page." before "GetImageWebPartAsync" on line: 191

PR #1017, opened 2022-11-04
With Get the context is still empty

StefanSchoof:
To use `context.Web.Lists.AsRequested()` you mus use `LoadAsync` and not `GetAsync`


jansenbe:
Nice catch @StefanSchoof , thanks for the PR!

Issue #897, opened 2022-06-24
Tell us what is missing in PnP Core SDK / what feature would you like to see added


jansenbe:
To further grow PnP Core SDK we'd like to understand what features you'd like to see added to PnP Core SDK. If you've need a feature in your project or you've suggestions for features to add then please create a new issue in our issue list so that we can track your ask.

valpvt:
The Author property is missing from the IWeb class.
This is not a lot but it would be quite useful to me for a reporting app I'm building.
Hope to see this property included in next release.

Thanks a lot for your great work !

gautamdsheth:
Feature request: 

Add capability to easily add Quick Links webpart on a page, similar to Image or Text webpart. 
Here, we will be able to define Link URL, Thumbnail Image or Icon , Alt text and maybe if possible order of showing links in the webpart.

I would be happy to add a new cmdlet like - Add-PnPPageQuickLinksWebpart or something like that in PnP PowerShell.

valpvt:
Feature request : 
Add capability to retrieve Private Channel Sites (like in Pnp PowerShell). 
Best regards

joshmiller1FEI:
with SharePointPnPCoreOnline I could connect Without a clientID 
    ```
 var authManager = new OfficeDevPnP.Core.AuthenticationManager();
            ClientContext ctx = authManager.GetWebLoginClientContext("https://example.sharepoint.com/sites/example");
            Web web = ctx.Web;
            ctx.Load(web);
            ctx.ExecuteQuery();
```

How can I achieve the above with the new pnp.core ?

The new Minimal Console Example here requires registration of an AAD App and subsequent Client ID.

thank you
Josh

jansenbe:
@joshmiller1FEI : the web login approach indeed is not in PnP Core SDK and that's by design, we're also not supporting Azure ACS anymore in PnP Core SDK. Both approaches are legacy authentication models, the only future proof authentication model for SharePoint is via OAuth and an Azure AD application.

fowl2:
@jansenbe Azure AD authentication still doesn't have feature parity with ACS. For example: 
* Site collection admins can't grant permissions, only tenant admins. 
* There's no UI for granting granular (site or list level) permissions. 
* Only certificates are supported, not secrets.

jansenbe:
[...]

PR #1021, opened 2022-11-06
Add ITermSet.GetTermsByCustomProperty method to retrieve a list of terms with a certain custom property key and value

mloitzl:
As discussed in https://github.com/pnp/pnpcore/issues/897#issuecomment-1303851341


mloitzl:
That was fast 🚀 Thanks @jansenbe  😎

Issue #1013, opened 2022-10-31
Docs for using Managed Identity auth


kkazala:
## Category

- [x] Documentation request

Would it make sense to add an example in [PnP Core Samples](https://pnp.github.io/pnpcore/demos/README.html) for using Azure Function with Managed Identity?
We need custom auth provider, I know =) 
I think it would be super nice to have a working example everyone could use.

I just implemented it, so if you are interested, I could contribute the sample. 
Or... is there something that is still not working, that I'm not yet aware?

jansenbe:
@kkazala : that would be an awesome addition! Happy to help along the process getting the PR in.

kkazala:
Perfect. Please give me a few days as I’m currently swamped. Latest next week

jansenbe:
PR merged, thanks again @kkazala 💪🚀👍

PR #1022, opened 2022-11-08
Az function managed identity

kkazala:
@jansenbe Could you have a look at the Managed Identity authentication sample (#1013 )?


kkazala:
@jansenbe awesome, thank you for accepting the PR! 😀 
Just realized that the sample doesn't show in TOC and the main readme for samples.
Should I update it as well?

jansenbe:
Already done that, thanks for the awesome sample!

PR #1037, opened 2022-11-23
Bugfix on modern team site creation. 

LeHailender:
If an empty or no description was passed the creation is crashed. It seems that the serialized payload for the graph call does not like empty strings.


jansenbe:
Nice catch @LeHailender !

Issue #1046, opened 2022-12-06
Support passing X509 object to OnBehalfOfAuthenticationProvider


rlv-dan:
## Category

- [x] Feature request

## Describe the feature

When creating an `X509CertificateAuthenticationProvider` you can either provide StoreName/StoreLocation/Thumbprint or a ready `X509Certificate2` object. Great! Another provider I was using is `OnBehalfOfAuthenticationProvider`. But this only accepts the StoreName/StoreLocation/Thumbprint parameters, not a certificate object.

_Originally posted by @rlv-dan in https://github.com/pnp/pnpcore/issues/897#issuecomment-1339067652_
## Describe the solution you'd like

Allow creation of `OnBehalfOfAuthenticationProvider` with an `X509Certificate2` object in addition to Thumbprint

mloitzl:
@jansenbe I could have a look at this one...

jansenbe:
Thanks for helping out @mloitzl 💪🚀

jansenbe:
Thanks for your PR @mloitzl ! @rlv-dan , you can use this using the next nightly build (version 1.8.33 or higher). Closing issue now, feel free to re-open in case needed

PR #1047, opened 2022-12-07
Support passing X509 object to OnBehalfOfAuthenticationProvider

mloitzl:
As required in https://github.com/pnp/pnpcore/issues/1046


src/sdk/PnP.Core.Auth.Test/Providers/OnBehalfOfAuthenticationProviderTests.cs
```diff
@@ -391,6 +434,30 @@ private static OnBehalfOfAuthenticationProvider PrepareOnBehalfOfAuthenticationP
             return provider;
         }
 
+        private static OnBehalfOfAuthenticationProvider PrepareOnBehalfOfAuthenticationProviderWithCertificateFile()
+        {
+            if (TestCommon.RunningInGitHubWorkflow())
```
mloitzl, line 439: @jansenbe I am not sure if it is possible to load the cert from test assets within a GitHub action

PR #1168, opened 2023-04-25
IconAlignment cannot be used to infer collapsibility

mmi-wp:
We can't rely on the IconAlignment property to figure out if the section is collapsible, because enabling and then disabling collapsiblity on a section will retain a value for the IconAlignment. Instead, look at the section type to figure out if the section should be expandable.


jansenbe:
@mmi-wp : thanks for this and the other PR's, good fixes!

PR #1179, opened 2023-05-12
Bugfix/replicate key and metadata catch all exceptions

cnesmark:
The ClientException in this catch will not actually catch any exception, and caused problems when NotImplementedException was thrown from BaseDataModel while attaching field links to a content type - even though the field was added.


jansenbe:
Thanks for sharing this @cnesmark . Do you have a code snippet that showed the problem so that I can build a repro + test case?

cnesmark:
Sure! It only fails the first time (upon fieldLink creation, that is). Updating the fieldLink works fine.

    private async Task EnsureContentTypeAndFieldAndFieldLink()
    {
      // Get site content types, also load the content types field links in one go
      await _context.ContentTypeHub.LoadAsync(
        p => p.Fields,
        p => p.ContentTypes.QueryProperties(
          p => p.Id,
          p => p.Name,
          p => p.Description,
          p => p.Group,
          p => p.StringId,
          p => p.FieldLinks.QueryProperties(
            p => p.Id,
            p => p.Name)
          )
        );

      var id = "0x0101009D1CB255DA76424F860D91F20E6C4118_AND_YOUR_ADDITIONAL_CTID_HERE";
      var name = "Your SitePage child";
      var group = "Your CT group";

      var contentType = _context.ContentTypeHub.ContentTypes.AsRequested().FirstOrDefault(p => p.StringId.Equals(id, StringComparison.InvariantCultureIgnoreCase));

      if (contentType == null)
      {
        _logger.LogInformation($"Adding ContentType [{id}] {name}");
        contentType = await _context.ContentTypeHub.ContentTypes.AddAsync(id, name, group: group);
      }
      else
      {
        _logger.LogInformation($"Updating ContentType [{id}] {name}");
        contentType.Name = name;
        contentType.Group = group;

        await contentType.UpdateAsync();
      }

      await contentType.FieldLinks.LoadAsync(
        p => p.FieldInternalName,
        p => p.Id,
        p => p.Name,
        p => p.ShowInDisplayForm,
        p => p.Hidden,
        p => p.Required
      );

      // Ensure the field
      var fieldId = "YOUR_FIELD_ID";
      var fieldInternalName = "Your_x0020_Internal_x0020_Name";
      var fieldDisplayName = "Your Display Name";
      var unlimitedLengthInDocumentLibrary = true;

      Guid guid = new(fieldId);
      var field = _context.ContentTypeHub.Fields.AsRequested().FirstOrDefault(f => f.Id == guid);

      field ??= await _context.ContentTypeHub.Fields.AddMultilineTextAsync(fieldInternalName, new FieldMultilineTextOptions()
      {
        Id = guid,
        RichText = false,
        UnlimitedLengthInDocumentLibrary = unlimitedLengthInDocumentLibrary
      });

      field.Group = "The Field Group";
      field.Title = fieldDisplayName;
      await field.UpdateAsync();

      // Ensure the field link
      var fieldLink = contentType.FieldLinks.AsRequested().FirstOrDefault(f => f.Id == field.Id);

      if (fieldLink == null)
      {
        _logger.LogInformation($"Attaching Field=[{field.InternalName}] to CT=[{contentType.Name}]");
        fieldLink = await contentType.FieldLinks.AddAsync(field, required: false);
      }
      await fieldLink.UpdateBatchAsync();

      // ...and maybe a few more other fieldLinks added/updated in the same way

      await _context.ExecuteAsync();
    }

jansenbe:
@cnesmark : quick question, in your code the `_context` points to a regular site and not to the content type hub site, right?

cnesmark:
@jansenbe Good question! No, the _context object in that code points to the content type hub. This is how it's created from my context wrapper class:


    public PnPContext GetClonedContentTypeHubContext()
    {
      var uri = new Uri(_context.Uri.AbsoluteUri.TrimEnd('/') + "/sites/contentTypeHub");
      _logger.LogInformation($"Cloning context for content type hub [{uri}]");
      return _context.Clone(uri);
    }

The particular reason why _context points to /sites/contenttypehub, is that it's impossible to access the PnPContext.ContentTypeHub property when the context points to the root site (i.e. "/" or just "") - as would be my preferred choice for this code (because that site always exists). The reason why root can't be used is in the ContentTypeHub constructor, line 22:

var request = api.ApiCall.Request.Replace(PnPContext.Uri.AbsolutePath, PnPConstants.ContentTypeHubUrl);

If PnPContext.Uri.AbsolutePath is an empty string, there will be no replacements. If the string is "/", you'll get something like https:/sites/contenttypehub/sites/contenttypehubyourtentantname.sharepoint.com/ and so on. (I have been thinking about submitting a fix for this as well, because I guess it should be possible to get to the CTH from root?)

jansenbe:
Thanks for the feedback @cnesmark . I think that since you're creating the context on the content type hub site you can workaround the issue by using `.Web.` versus `.ContentTypeHub.`.

The fact that this fails when the content is created on the root site collection is a bug, just like my finding (root cause there is that the id's of the context are used to build the CSOM payload instead of the id's of the content type hub site).

jansenbe:
@cnesmark : I pushed a fix that should allow using the root site collection in combination with `.ContentTypeHub.`

jansenbe:
@cnesmark : I've pushed various fixes, this PR is not needed. Please retest with the next nightly and if you've still issues then let me know via an issue

PR #1183, opened 2023-05-19
Add SectionType property of CanvasSection to public model

wilecoyotegenius:
This PR adds SectionType property to ICanvasSection interface, so the property may be utilized in public model.
I am working on a PR for pnpframework, extending PNP Provisioning to support collapsible sections. In order to properly implement it, I need to extend ICanvasSection with SectionType first (as pnpframework utilizes that interface).


jansenbe:
@wilecoyotegenius : the latest PnP Framework provisioning schema (version 2022 09) supports collapsible sections already, a bit confused on what you're trying to do. Are you saying the current support in PnP Framework is not working as expected?

wilecoyotegenius:
@jansenbe Apologies. Just noticed the template I used is using an old schema.
Sorry for taking your time.

PR #1207, opened 2023-06-15
fix(role assignments): incorrect url for getting by principal id

JakeStanger:
Fixes the incorrect call on role assignments when getting by principal id:

```cs
_api/web/roleAssignments/getByPrincipal(10)
```

to the correct:

```cs
_api/web/roleAssignments/getByPrincipalId(10)
```


jansenbe:
Nice find @JakeStanger , appreciate the PR with the fix. There indeed is no "GetPrincipal" method on the RoleAssignments collection in SP REST.

PR #1269, opened 2023-09-09
Add Task overload of constructor to external provider to allow it to run async

quails4Eva:
Allow passing a Task to the constructor so auth can be done async. Some of the complexity is to ensure it's backwards compatible with existing code.

I don't think the tests are quite right yet, would appreciate some guidance on how best to handle it for this project. Simplest thing to do would be to copy the existing ones but just use the new constructor.

A few things I was thinking about.
- Do the tests actually need to connect to Sharepoint? Would unit tests be sufficient? Perhaps this would be inconsistent with the other areas of the project.
- Constructor could take an interface rather than a Func?
- No use of mocking library? Not sure if you can actually mock a Func.
- In tests could use data attribute to control which constructor to use and whether to use fake version of token provider. Possibly a bit overkill.


jansenbe:
@quails4Eva : thanks for looking into this, being able to invoke the external access token retrieval code asynchronous does make sense.

PR #1387, opened 2024-02-08
Ensure the finishupload endpoint is always called

alexchx:
In rare condition, the stream length can be `N` times of the `chunkSizeBytes`, then the `bytesRead` of the last read will be 0 and jump out of the while loop, as a result, the upload wouldn't be finalized by this endpoint `finishupload`, and leave a broken file in SharePoint.


jansenbe:
Thanks for submitting this fix @alexchx , I'll process your PR in the coming days (just back from a break)

Issue #1352, opened 2024-01-16
Sensitivity label not being assigned when creating team site


bricenocar:
## Category

- [x] Bug

## Describe the bug
I am using the following options for team site creation:

var teamSiteToCreate = new TeamSiteOptions(StringUtilities.NormalizeSiteAlias(alias), title)
{                    
    Description = description,
    Language = language,                    
    SensitivityLabelId = sensitivityLabel,
    WelcomeEmailDisabled = true,
    IsPublic = isPublic
};

using var newSiteContext = await context.GetSiteCollectionManager().CreateSiteCollectionAsync(teamSiteToCreate, siteCreationOptions)

The site does not get the sensitivity label id set. However, it functions seamlessly if I create a communication site.

## Expected behavior

The site gets the label assigned

## Environment details
- **SDK version**: PnP.Core (1.10.0)
- **OS**: Windows 11

### Additional context

I did a quick check and it looks like the error might be in the file: pnpcore\src\sdk\PnP.Core.Admin\Model\SharePoint\Core\Internal\SiteCollectionCreator.cs
In line: 255. The graph group creation option is being set with the key "SensitivityLabel", maybe it has been udated/changed in graph.

jansenbe:
Thanks for reporting this @bricenocar. Would it be possible use Fiddler and create a site in the UI using a sensitivity label and then check for the payload used to create the site. This will validate you're using the correct label id. I assume (did not cross check) that the UI is calling a REST API, we however are calling a Graph API. 

In your code, are you using application permissions or delegated?

bricenocar:
Hello @jansenbe,

I have done some tests on both a team site (with group) and a communication site (without group). The creation of a communication site via the SharePoint (SP) REST API proceeds smoothly, and the sensitivity label is assigned without any issues.

However, the creation of a team site involves a slightly different process. During site creation, the group is assigned the sensitivity label via Microsoft Graph. I verified this by inspecting the Graph request in the local debugging session of my Azure Function. Despite the successful Graph request, the sensitivity label is not assigned.

To troubleshoot, I attempted to create the site using the Graph Explorer with a POST request to https://graph.microsoft.com/v1.0/groups. This time, I included the sensitivity label in the payload:

```
{
    "displayName": "Just a test",
    "mailEnabled": true,
    "mailNickname": "library",
    "securityEnabled": true,
    "groupTypes": [
        "Unified"
    ],
    "assignedLabels": [
        {
[...]

PR #1389, opened 2024-02-12
Add code to set sensitivity label to a group via updating associated site properties

wilecoyotegenius:
As per #1352, when creating a team site using TeamSiteOptions, sensitivity label is not getting correctly applied (it shows correctly in SPO Admin, but does not in Azure Portal nor in Graph API when querying the created group).

This PR modifies the team site creation process, by removing sensitivity label from 'creationOptions' and applying label to the group via updating associated site properties just after group creation.


jansenbe:
Thanks for this fix @wilecoyotegenius ! I'll process your PR in the coming days, I'm just back from a break

jansenbe:
@wilecoyotegenius: this change was merged, I added the needed test cases to ensure we cover this scenario.

@bricinocar: FYI

wilecoyotegenius:
@jansenbe Glad I could help :)

bricenocar:
Thanks a lot @jansenbe !

Issue #1369, opened 2024-01-31
LoadListDataAsStream with Location column: System.FormatException: One of the identified items was in an invalid format.


wizofaus:
## Category

- [x] Bug

## Describe the bug

Exception thrown calling LoadListDataAsStream(Async) when the document library has location columns and rows with data for such columns.

System.FormatException: One of the identified items was in an invalid format.
         at System.Text.Json.ThrowHelper.ThrowFormatException()
         at System.Text.Json.JsonElement.GetInt32()
         at PnP.Core.Model.SharePoint.ListDataAsStreamHandler.GetJsonPropertyValueAsString(JsonElement propertyValue)
         at PnP.Core.Model.SharePoint.ListDataAsStreamHandler.TransformRowData(JsonElement row, IFieldCollection fields, Dictionary`2 fieldLookupCache)
         at PnP.Core.Model.SharePoint.ListDataAsStreamHandler.Deserialize(String json, List list)
         at PnP.Core.Model.SharePoint.List.LoadListDataAsStreamAsync(RenderListDataOptions renderOptions)
         at TDI.DocumentService.SPLibraryConnector.GetItems(ItemsParameters itemsParameters, Boolean asUser) in

## Repro steps

Create a document library with a location column and add at least one document with a valid location for said column.

Call as per:

```
                var output = await lib.LoadListDataAsStreamAsync(new RenderListDataOptions()
                {
                    RenderOptions = RenderListDataOptionsFlags.None,
                    MergeDefaultView = true,
                    AddRequiredFields = true,
                }).ConfigureAwait(false);
```
 

## Expected behavior

No exception should be thrown, and location data available in item values.

## Environment details (*development & target environment*)

- **SDK version**: 1.11.0
- **OS**: Windows 10 
- **SDK used in**: ASP.Net (Core) Web app 
- **Framework**: .NET Core v8.0
- **Browser(s)**: N/A
- **Tooling**: Visual Studio 2022
- **Additional details**: *The more context you can provide, the easier it is (and therefore quicker) to help.*

### Additional context

*Thanks for your contribution! Sharing is caring.*

wizofaus:
Due to attempting to parse decimal places as integers.   GetJsonPropertyValueAsString( ) shouldn't assume ValueKind == Number means integer (I tried GetDecimal(), but GetDouble() should work for 99.99% of cases and IFieldLocationValue uses double for Latitude/Longitude anyway).

wizofaus:
[...]

Issue #1365, opened 2024-01-30
LoadListDataAsStreamAsync: System.InvalidOperationException (The requested operation requires an element of type 'String') when view contains Image field data


wizofaus:
## Category

- [x] Bug

## Describe the bug

Exception thrown calling LoadListDataAsStream(Async) when the document library has image columns and rows with data for such columns.
The data returned by the SPO REST API RenderListDataAsStream looks like:
```
{"fileName":"filename.jpg","serverRelativeUrl":"/sites/TestUser/SiteAssets/Lists/eda0a08f-8db5-4ad3-81d6-cc2e66357fa3/filenamejpg","id":"1b813e01-5d0d-4c2e-8eca-a72e718d61ce","serverUrl":"https://tenant.sharepoint.com","thumbnailRenderer":{"spItemUrl":"https://tenant.sharepoint.com:443/_api/v2.1/drives/...","fileVersion":1,"sponsorToken":"..."}
```
But PnP.Core (ListDataAsStreamHandler.Deserialize) is trying to parse it as a string.

## Repro steps

Create a document library with an image column and add at least one document with a valid image for said column.

Call as per:

```
                var output = await lib.LoadListDataAsStreamAsync(new RenderListDataOptions()
                {
                    RenderOptions = RenderListDataOptionsFlags.None,
                    MergeDefaultView = true,
                    AddRequiredFields = true,
                }).ConfigureAwait(false);
```
 

## Expected behavior

No exception should be thrown, and image data available in item values.
(I'd also say, if there really *were* an issue parsing the JSON response, I wouldn't expect it to be an InvalidOperationException...)

## Environment details (*development & target environment*)

- **SDK version**: 1.11.0
- **OS**: Windows 10 
- **SDK used in**: ASP.Net (Core) Web app 
- **Framework**: .NET Core v8.0
- **Browser(s)**: N/A
- **Tooling**: Visual Studio 2022
- **Additional details**: *The more context you can provide, the easier it is (and therefore quicker) to help.*

### Additional context

*Thanks for your contribution! Sharing is caring.*

wizofaus:
Happy to contribute a fix for this if needed, how long would it take typically to get included in the next NuGet release?  Though if someone can see a workaround, even if it's just to ensure such data is ignored and not returned (I can live without it for now), that'd be good.

wizofaus:
[...]

PR #1411, opened 2024-03-02
#1365 #1369 prevent crash on image and location columns

wizofaus:
unit test will only work with mocked data or if you explicitly create list on your site with correct columns


src/sdk/PnP.Core/Model/SharePoint/Core/Internal/FieldThumbnailValue.cs
```diff
@@ -0,0 +1,94 @@
...
+            {
+                ThumbnailRenderer = properties["thumbnailRenderer"];
+            }
+            // Clear changes
+            Commit();
+
+            return this;
+        }
+
+        internal override object ToJson()
```
wizofaus, line 73: incomplete as I'm not sure when this is used

src/sdk/PnP.Core/Model/SharePoint/Core/Internal/ListDataAsStreamHandler.cs
```diff
@@ -692,6 +693,10 @@ private static object GetJsonPropertyValue(JsonElement propertyValue, FieldType
                         {
                             return null;
                         }
+                        else if (propertyValue.ValueKind == JsonValueKind.Object)
+                        {
+                            return propertyValue.ToObject<System.Dynamic.ExpandoObject>();
```
wizofaus, line 698: this is a "just in case" for other field types that happen to be treated as objects

PR #1464, opened 2024-05-07
allow access to the SPRequest Guid in a BacthRequest

StefanSchoof:
To help my use case from #1460


jansenbe:
Thanks @StefanSchoof! Given I'm travelling this week and still catching up on things from last week off processing this PR will probably happen next week.

jansenbe:
@StefanSchoof : also see https://pnp.github.io/pnpcore/using-the-sdk/basics-responseheaders.html

Issue #1461, opened 2024-05-06
Visibility HiddenMembers and custom properties creating a Team Site


danielpastoor:
## Category

- [x] Feature request

## Describe the feature

My feature request is that a Team Site with group also can have the visibility HiddenMembers instead of only public and private.

The enum already supports the value but the model needs to change I think for TeamSiteCreation because of the IsPublic property.

Also it would be handy if we could add custom properties to the TeamSiteWithGroup Creation model. Like a dictionary called additionalData to support as example Education Class groups/Teams. Then you need to add a custom property to the group before creating the Teams to it.

jansenbe:
Closing as the PR is merged. Thanks @danielpastoor

PR #1467, opened 2024-05-14
#1461 Add better support for the Visibility HiddenMembership and added additionaldata support for creating a Group

danielpastoor:
#1461 Add better support for the Visibility HiddenMembership when creating a site with group.

And fixed an issue when loading team data if the Visibility is HiddenMembership by adding it to the TeamVisibility.cs enum. And last added support for additional creation properties when creating a Group like the property for creating an Education Class Group.


jansenbe:
Thanks @danielpastoor! Given I'm travelling this week and still catching up on things from last week off processing this PR will probably happen next week.

PR #1468, opened 2024-05-14
Added specialization support when adding a Team to a group

danielpastoor:
Added specialization support when adding a Team to a group


jansenbe:
Thanks @danielpastoor! Given I'm travelling this week and still catching up on things from last week off processing this PR will probably happen next week.

Issue #1497, opened 2024-07-13
[Feature] SharePoint page sections - collapsible sections


mkm17:
## Category

- [x] Feature request

## Describe the feature

Add collapsible options to SharePoint page sections.

## Additional context

Some time ago, I proposed extending the PnP.Powershell [Add-PnPPageSection](https://github.com/pnp/powershell/issues/4008) cmdlet to include collapsible section options.

I have reviewed the current [capabilities in PnP.Core](https://pnp.github.io/pnpcore/using-the-sdk/pages-intro.html#configuring-page-sections), and it appears that the existing code allows setting it up. However, the effect is not visible on the page.

During my tests, it seems that making the changes in [my branch](https://github.com/mkm17/pnpcore/tree/page_section_options) would be enough to achieve the desired effect.

I'm new to contributing to PnP.Core. Could you please guide me on what I should do next?
Have you encountered similar issues with collapsible sections?

Looking forward to receiving any tips.

jansenbe:
@mkm17 : please submit your updates as a PR, I can then test and merge.

jansenbe:
@mkm17 : as mentioned in your PR in my testing this just works. Anyhow, feel free to further discuss this via the still open PR.

PR #1499, opened 2024-07-26
Update Page.cs and CanvasColumn.cs to handle collapsible sections correctly

mkm17:
As mentioned in issue #1497, the following PR contains a fix for collapsible sections in SharePoint pages.

### DIFF
--- a/src/sdk/PnP.Core/Model/SharePoint/Pages/Internal/CanvasColumn.cs
+++ b/src/sdk/PnP.Core/Model/SharePoint/Pages/Internal/CanvasColumn.cs
@@ -151,6 +151,18 @@ namespace PnP.Core.Model.SharePoint
                     }
                 };
 
+                if (Section.Collapsible)
+                {
+                    clientSideCanvasPosition.ZoneGroupMetadata = new SectionZoneGroupMetadata()
+                    {
+                        IsExpanded = Section.IsExpanded,
+                        DisplayName = Section.DisplayName,
+                        ShowDividerLine = Section.ShowDividerLine,
+                        IconAlignment = Section.IconAlignment == IconAlignment.Right ? "right" : "left",
+                        Type = 1,
+                    };
+                }
+
                 var jsonControlData = JsonSerializer.Serialize(clientSideCanvasPosition);
 
                 html.Append($@"<div {CanvasControlAttribute}="""" {CanvasDataVersionAttribute}=""{DataVersion}"" {ControlDataAttribute}=""{jsonControlData.Replace("\"", "&quot;")}""></div>");
--- a/src/sdk/PnP.Core/Model/SharePoint/Pages/Internal/Page.cs
+++ b/src/sdk/PnP.Core/Model/SharePoint/Pages/Internal/Page.cs
@@ -1235,6 +1235,8 @@ namespace PnP.Core.Model.SharePoint
                             }
                         }
 
+                        ApplyCollapsibleSectionSettings(sectionData.ZoneGroupMetadata, currentSection);
+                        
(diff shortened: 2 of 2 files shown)


jansenbe:
Hey @mkm17 ,

I've just verified that our collapsible section test is working as expected (https://github.com/pnp/pnpcore/blob/dev/src/sdk/PnP.Core.Test/SharePoint/PagesTests.cs#L828-L910). Can you first verify that this is not working for you?

mkm17:
Hi @jansenbe, sorry for the delay in response. I've been quite busy lately.

In my testing, I tried adding some collapsible sections to the same page. The issue appears when there's already a collapsible section on the page. and I tried to add another one, I noticed that all properties related to the collapse function disappeared from the properties. 

image

For this page, I get this result ->

image

Then, when we save the page, all previously set collapsible sections are, of course, gone. Maybe this issue occurs only with empty sections, but I still think we should address it.

In the end, the result of the provided code is three empty sections.
image

I hope that I haven't overlooked anything, like the version or any obvious settings. Let me know if you can reproduce the same behavior.

My changes should resolve this issue, but since I'm not fluent with this repo, I hope they were added correctly.

jansenbe:
@mkm17 : can you verify that when then sections have content the same problem occurs?

mkm17:
Hi @jansenbe, the sections with web parts are fine.

The page before adding the section:
image

The page after adding the section (the section without web parts has been cleared):
image

If you need any further research, I’m here to help!

jansenbe:
@mkm17 : so the only issue is only with empty sections...not sure if we really need this as empty sections collapsed or not collapsed is almost the same

mkm17:
@jansenbe, hi, and thank you for your comment. The same thing happens when saving an empty collapsible section, so this code without control inside does not work (at least in my case): https://pnp.github.io/pnpcore/using-the-sdk/pages-intro.html#configuring-page-sections.

To add a little bit of background on why I am pushing for this change, I wanted to extend this command in PnP PowerShell: https://pnp.github.io/powershell/cmdlets/Add-PnPPageSection.html. Without the possibility to add a collapsible section to an empty section, it will be impossible to extend this command.

If adding the next empty collapsible section resets all previous empty ones, then it won't be possible to add more empty sections using these commands in the future:
```
Add-PnPPageSection -Page "Test" -SectionTemplate TwoColumn -IsCollapsibleSection -IconAlignment Right -DisplayName "Section PnP7"
Add-PnPPageSection -Page "Test" -SectionTemplate TwoColumn -IsCollapsibleSection -IconAlignment Right -DisplayName "Section PnP8"
```

Of course, it is not a crucial change, so if it is better not to touch this part of the code in PnP Core, then I will just extend the PnP PowerShell commands like Add-PnPPageWebPart, which should work with the current setup. :)

czullu:
a empty section seems to look like this in 'data-sp-controldata':
{"position":{"layoutIndex":1,"zoneIndex":2,"zoneId":"7cac6f3c-c012-47a4-89ab-cbe3611349d5","sectionIndex":1,"sectionFactor":6,"controlIndex":1},"emphasis":{"zoneEmphasis":2},"zoneGroupMetadata":{"type":1,"isExpanded":false,"showDividerLine":false,"iconAlignment":"right","isTitleEnabled":false},"id":"emptySection","controlType":1,"addedFromPersistedData":true}

CanvasControl.GetType(controlData) can not get the Type... so we not read it to the model.
=>Working on it

czullu:
should be solved in  https://github.com/pnp/pnpcore/pull/1682

jansenbe:
Thanks @czullu ! I'll close this PR then @mkm17

PR #1058, opened 2022-12-14
The JSON value could not be converted to System.Int32 : Handle Pnp Provisioning template extraction failure from Project Management site template

stokuri:
[Cant extract Pnp Provisioning template from Project Management site template](https://github.com/pnp/powershell/discussions/2421)

Get-PnPSiteTemplate fails with the following error.

`The JSON value could not be converted to System.Int32. Path: $.position.sectionIndex | LineNumber: 0 | BytePositionInLine: 107.`

The issue is in [pnpcore](https://github.com/pnp/pnpcore) while Deserializing the [controlDataJson](https://github.com/pnp/pnpcore/blob/dev/src/sdk/PnP.Core/Model/SharePoint/Pages/Internal/CanvasControl.cs#L331).

The sample JSON for which the failure occurs:

`controlDataJson: {"controlType":4,"id":"00000000-0000-0000-0000-100000000005","position":{"zoneIndex":1.0,"sectionIndex":2.0,"controlIndex":2.0,"sectionFactor":6,"layoutIndex":1},"emphasis":{}}`

`controlDataJson: {"controlType":3,"id":"00000000-0000-0000-0000-100000000006","position":{"zoneIndex":1.0,"sectionIndex":2.0,"controlIndex":3.0,"sectionFactor":6,"layoutIndex":1},"webPartId":"8654b779-4886-46d4-8ffb-b5ed960ee986","emphasis":{},"addedFromPersistedData":true}`

The parser is failing to read `"sectionIndex":2.0` property which is defend as `int` in [CanvasPosition.cs](https://github.com/pnp/pnpcore/blob/dev/src/sdk/PnP.Core/Model/SharePoint/Pages/Internal/CanvasPosition.cs#L21) but the actual value in JSON is `float`

Issue #1043, opened 2022-11-29
Object reference not set to an instance of an object - IAuthenticationProvider dependency injection - Azure Function


stonekw:
## Category

- [x] Bug

## Describe the bug

I have an azure function that will trigger some action against a site specified by URL in the function query string. In order to do this, I am trying to inject the IAuthenticationProviderFactory and IPnPContextFactory into my function code to create the PnPContext on the fly when the function is called (we don't know the site URL until the function is called). null reference exception occurs when I try 

` IAuthenticationProvider authProvider = _authFactory.Create(targetEnv.AuthConfigName);`

In this snippet, _authFactory is not null, nor is targetEnv or AuthConfigName. The exception appears to be internal to the Create() method

Here is my startup.cs:

`namespace MyNamespace.Functions
{
    public class Startup : FunctionsStartup
    {
        public override void Configure(IFunctionsHostBuilder builder)
        {
            builder.Services.AddLogging();

            // Add the PnP Core SDK library services
            builder.Services.AddPnPCore(options =>
            {
                // Disable telemetry because of mixed versions on AppInsights dependencies in AZ functions
                options.DisableTelemetry = true;

                options.PnPContext.GraphFirst = true;
                options.HttpRequests.UserAgent = "NONISV|MyNamespace|SecurityMatrixFunction";
                options.HttpRequests.SharePointRest = new PnPCoreHttpRequestsSharePointRestOptions
                {
                    UseRetryAfterHeader = true,
                    MaxRetries = 5,
                    DelayInSeconds = 3,
                    UseIncrementalDelay = true,
                };
            });

            // Add the PnP Core SDK Authentication Providers
            builder.Services.AddPnPCoreAuthentication(options =>
            {

                foreach (MyPnPEnvironment env in AppConfig.Environments)
                {
                    //pnpcore getting a dictionary error if the config duplicates AuthConfigNames, added this check. 
                    if (options.Credentials.Configurations.ContainsKey(env.AuthConfigName) == false)
                    {
                        //PFX cert is in keyvault and lastpass. Password for certificate is in LastPass. 
                        //using managed identity DefaultAzureCredential to access Key Vault only. The same certificate is used to access field tenant and HQ tenant, both while debugging locally and published as a prod function 
                            CertificateClient client = new CertificateClient(vaultUri: new Uri(env.HQTenantKeyVaultUri), credential: new DefaultAzureCredential());
[...]

PR #1543, opened 2024-09-26
Fix check for HeaderTypes as new PageTitle WebPart in OneColumnFullWIth is not always in first section

czullu:
The original fix was expecting the PageTitle WebPart to be the first in the first section which needs to be a OneColumnFullWith.
#1058
#1043

It seems that this assumption was wrong and it can be in any section. The check has been changed to reflect this.
There is a related fix for pnp.framework as the check is present there as well.


magarma:
In some templates, PageTitle is not comming with OneColumnFullWith
image

Do you know a workaround for these cases?
@czullu  @jansenbe

PR #1563, opened 2024-10-26
Fix #1562

### LINKED ISSUE #1562: Querying for pages fails when Site Pages library is missing
## Category

- [x] Bug

## Describe the bug

https://github.com/pnp/powershell/issues/3959 is caused when creating a Page with Add-PnpPage but it fails in querying the pages because the site is missing 

## Steps to reproduce

See this issue for details:
https://github.com/pnp/powershell/issues/3959

## Expected behavior

If trying to create a Modern page on a site, it's expected "Site Pages" library exists or is created if it's missing.

## Other stuff

I'll submit a PR for adjusting, please review.

koskila:
Makes sure EnsurePagesLibrary ensures pages library for sites where "Site Pages" is not activated by activating the feature.

This is to address https://github.com/pnp/powershell/issues/3959

### DIFF
--- a/src/sdk/PnP.Core/Model/SharePoint/Pages/Internal/Page.cs
+++ b/src/sdk/PnP.Core/Model/SharePoint/Pages/Internal/Page.cs
@@ -472,7 +472,7 @@ namespace PnP.Core.Model.SharePoint
             return new Page(context, pagesLibrary, null, pageLayoutType);
         }
 
-        private static async Task<IList> EnsurePagesLibraryAsync(PnPContext context)
+        private static async Task<IList> EnsurePagesLibraryAsync(PnPContext context, bool firstTry = true)
         {
             IList pagesLibrary = null;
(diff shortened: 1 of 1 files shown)


jansenbe:
@koskila : I would not just change the default behavior here as for example automatically enabling a SitePages library on a classic publishing portal is not the right thing to do in my mind. Can you change the PR like this:
1. Update the name of the private EnsurePagesLibraryAsync method to EnsurePagesLibraryImplementationAsync
2. Add a new EnsurePagesLibraryAsync method that calls the EnsurePagesLibraryImplementationAsync + adds your enhancement. This method can be made a public one by updating the IPages interface

This way the internal flow is not changed and we don't automatically create a SitePages library from existing code paths, if someone however wants this behavior then can call the public EnsurePagesLibraryAsync  method

jansenbe:
Closing per no activity

PR #1598, opened 2025-01-15
The Url for Root-Site-API Request had // in path

czullu:
The Url combine in ListItem, FieldLinks, Files did not take root-site-url into account


jansenbe:
Nice catch @czullu , thanks a lot!

PR #1620, opened 2025-02-28
#1548 - feature: recycle folder

mloitzl:

PR #1621, opened 2025-02-28
Feature/recycle folder less duplication

mloitzl:


mloitzl:
The same as https://github.com/pnp/pnpcore/pull/1620 but with less duplication.

Introducing RecyclableBaseDataModel<TModel> feels strange to reduce duplicated code, since all ExpandoBaseDataModel<TModel> inherit Recycle capabilities, but only ListItem actually is able to be recycled.

I am not sure that there are others than File, Folder and List having recycle capabilities that could be inherited.

I don't know, @jansenbe should I further elaborate on that?

Actually I am not a fan of avoiding duplication at the cost of readability.

jansenbe:
@mloitzl : let's favor readability, I'll merge your other PR.

PR #1636, opened 2025-04-08
prevent blank lines when html tags have attributes

robi26:
When updating a page containing html tags with attributes the html was unnecessary wrapped with p tags.
E.g. ``<p style="text-align:justify;">...</p>`` was wrapped in p tags again what lead to blank lines.

This PR uses a regex to check a valid html tag is present and prevents the wrapping with p tags.


jansenbe:
Nice improvement @robi26 , thanks for your contribution!

PR #1658, opened 2025-06-10
Improved page parsing robustness when wrongly decoded quotes

singingknight:
Handle situation where pages contains WebParts with properties, which when HtmlDecoded generates invalid JSON giving errors like:

```
'E' is invalid after a value. Expected either ',', '}', or ']'. Path: $ | LineNumber: 0 | BytePositionInLine: 17277.
```

Tries to replace `"` with `\"`, if this fails just export with basic info and title set to `Failed to deserialize WebPart`


PedroMordeP:
Hello, thank you @singingknight for this, but for my case it does not escape all the quotes and then ignores the web part
I will try to do more tests for my case if I can improve something

PR #1659, opened 2025-06-10
Feature/allow optional configuration of SharePointRestClient and MicrosoftGraphClient

kola-tm:
added optional configuration to be applied to SharePointRestClient and MicrosoftGraphClient. This will allow delegating handlers from a consumer of this SDK to be applied to either or both clients


jansenbe:
Thanks @kola-tm , nice work!

Question: can you also create a PR that adds a brief chapter to https://github.com/pnp/pnpcore/blob/dev/docs/using-the-sdk/basics-settings.md describing this new configuration option?

kola-tm:
@jansenbe Thanks for completing the PR. What are the time scales for next release?

And yes I am happy to update the documentation.

jansenbe:
We've not locked timing for a next release, the nightly builds however are very stable as well

PR #1674, opened 2025-07-09
Escape JSON from the Property value

PedroMordeP:
Fix issue `'a' is invalid after a value. Expected either ',', '}', or ']'`

This fix is to find and escape a JSON string value inside a property with HTML content with data attributes with JSON

example: 
escape JSON snippet: { "k1":"v1", "k2":"{v2}", "k3": ["k4":"v4","k5":"v5"] } from "TemplateHtml": ".... class=\\"portal-search-page\\" data-config-json=\\"{ "k1":"v1", "k2":"{v2}", "k3": ["k4":"v4","k5":"v5"] }\\" ) ...


PedroMordeP:
@singingknight I created this pull request to escaping JSON object inside a HTML property.

Your code only puts the escape in the first 3 double quotes then the error is at the end of the JSON string}\" and don't escape all the double quotes after the 3rd and then ignores the web part since it can resolve the issue.

Example:
 data-config-json=\\"{\\"apibaseuri\\":\\"https://mysite.com","groupid":"{sitecollectionconnectedoffice365groupid}","detailsPagePath":"","showViews":false,"view":"columns"}\"

Not sure if what I did can be included in your code or improved.

Related to https://github.com/pnp/pnpcore/pull/1658

PR #1676, opened 2025-07-12
Add DeviceCodeVerification action to DeviceCode options

ymasaoka:
Introduces a DeviceCodeVerification property to PnPCoreAuthenticationDeviceCodeOptions, allowing an Action<DeviceCodeNotification> to notify end users about device code requests.


jansenbe:
@ymasaoka : why do we need this? We already have https://github.com/pnp/pnpcore/blob/dev/src/sdk/PnP.Core.Auth/Public/DeviceCodeAuthenticationProvider.cs#L27 and https://github.com/pnp/pnpcore/blob/dev/src/sdk/PnP.Core.Auth.Test/Providers/DeviceCodeAuthenticationProviderTests.cs shows how this can be used.

jansenbe:
closing per no response. Feel free to open an issue if the explained approach is not working

Issue #1686, opened 2025-09-03
Support for HeadingLevel in collapsible sections


shunsaker:
## Category

- [x] Bug

## Describe the bug

pnpcore can only create collapsible section headers as `<h2>`, but sharepoint UI supports h2, h3, and h4. Additionally, loading a page into pnpcore and saving it will cause all collapsible section headers to be reset to h2

I can't upload screenshots, here's what I'm talking about:
https://support.microsoft.com/images/en-us/e23a8ced-7745-4279-8949-4fb272036c91

## Steps to reproduce

1. Create a new page
1.1 Add a collapsible section
1.2 Set "Choose title heading level" to "Heading 4"
1.3 Save page
1. Load page in pnpcore
1. Save page in pnpcore
1. Load sharepoint page on sharepoint and see "heading level" has returned to "Heading 2" 

## Expected behavior

Users should be able to read, configure, and save collapsible section heading level

## Environment details (*development & target environment*)

PnP.Core version 1.15
.NET 8

### Additional context

Looking at the save request when changing collapsible section heading levels through the browser, it looks like `headingLevel` is an int property in `zoneGroupMetadata`. I've created a PR that should work. See #1687

shunsaker:
I manually tested my changes and they do in fact work

jansenbe:
Thanks for the pr @shunsaker , I today merged it

PR #1687, opened 2025-09-03
Support for HeadingLevel in collapsible sections

shunsaker:
Support for HeadingLevel in collapsible sections
See  #1686

### DIFF
--- a/src/sdk/PnP.Core/Model/SharePoint/Pages/Internal/CanvasSection.cs
+++ b/src/sdk/PnP.Core/Model/SharePoint/Pages/Internal/CanvasSection.cs
@@ -204,6 +204,12 @@ namespace PnP.Core.Model.SharePoint
         /// </summary>
         public string DisplayName { get; set; }
 
+
+        /// <summary>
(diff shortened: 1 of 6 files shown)


jansenbe:
@shunsaker : nice work!

Question: if the default is 2, I think it's also needed to ensure HeadingLevel value is initialized to 2. We cannot assume every page has this value in it's json payload + we need to support creating pages where developers don't set the heading level

shunsaker:
> @shunsaker : nice work!
> 
> Question: if the default is 2, I think it's also needed to ensure HeadingLevel value is initialized to 2. We cannot assume every page has this value in it's json payload + we need to support creating pages where developers don't set the heading level


@jansenbe 
I thought about that. When it is empty sharepoint makes it 2, so it seemed okay to leave it empty and tell Json not to serialize it if it is 0. Which is exactly what it does today because it can't set that property

jansenbe:
Thx @shunsaker : I'll do some testing with the PR

shunsaker:
@jansenbe Bump. Have a minor feature blocked by this

PR #1705, opened 2025-11-14
Update workflows and projects to .NET 10.0 and upgrade dependencies

gautamdsheth:
PR for .NET 10 update.
Had to explicitly mention the class due to some breaking changes in .NET 10.


luismanez:
Any timeline here? we must upgrade our project to .NET 10. If we can help with whatever, happy to do it.
/cc @jansenbe 
Many thanks!

PR #1744, opened 2026-04-03
Retrieval API implementation

### LINKED ISSUE #1741: MS Graph API Copilot API Support
## Category

- [x] Feature request

## Describe the feature

We should support Copilot API in the SDK
https://learn.microsoft.com/en-us/microsoft-365/copilot/extensibility/api/ai-services/retrieval/overview

## Describe the solution you'd like

I would love to be able to call the Copilot API from the SDK and later from PnP.PowerShell

mgwojciech:
Added support for using MS Graph API Copilot Retrieval API
#1741


pkbullock:
Hey @mgwojciech, can you expand to include the docs for this please? To give examples of how to use and permissions, id use other areas as an example of documentation standard.

mgwojciech:
Sure thing. I wanted to verify first if everything is technically sounds. If that the case I will add the docs

pkbullock:
Awesome stuff! I have been testing this.

For the retrieval API, this looks good and can retrieve results (once I remembered to setup the tests lol).

The test when in non-mocking setup, would only work though for your setup (mocking will be fine of course), just wondering if you should add comments or code to setup the test for the next person working on this can ensure it passes when not mocking. 

There is a pattern of sync/async methods in other endpoints we have covered e.g. SharePoint, wondering if this too should align too?

src/sdk/PnP.Core/Model/Copilot/Public/ICopilot.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// </summary>
+    [ConcreteType(typeof(Internal.Copilot))]
+    public interface ICopilot
+    {
+        /// <summary>
+        /// The Microsoft 365 Copilot Retrieval API allows for the retrieval of relevant text extracts from SharePoint, OneDrive, and Copilot connectors content that the calling user has access to, while respecting the defined access controls within the tenant. Use the Retrieval API to ground your generative AI solutions with Microsoft 365 data while optimizing for context recall.
+        /// </summary>
+        /// <param name="request"></param>
+        /// <returns></returns>
+        public Task<RetrievalResponse> Retrive(RetrievalRequest request);
```
Adam-it, line 18: Wouldn't it be better if we use `RetrieveAsync`
Like simiarly we do for `SendMailAsync` or `SearchAsync` or `RawRequestAsync`

src/sdk/PnP.Core/Model/Copilot/Internal/Copilot.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Text.Json;
+using System.Threading.Tasks;
+
+namespace PnP.Core.Model.Copilot.Internal
+{
+    /// <summary>
+    /// Provides methods for interacting with Microsoft Graph services to perform retrieval operations using a specified
+    /// Microsoft Graph client.
+    /// </summary>
+    internal sealed class Copilot : BaseDataModel<ICopilot>, ICopilot
```
Adam-it, line 17: Shouldn't we decorate it with something like `[GraphType]` ?

Adam-it (commented):
@mgwojciech awesome work I only left two tiny comments and I guess we should be ready to merge 👍

src/sdk/PnP.Core/Model/Copilot/Internal/Copilot.cs
mgwojciech: Copilot is not a type on it's own - to my best understanding it's just a part of the path - not sure if we can define it as a GraphType (but correct me if I'm wrong here)

PR #1797, opened 2026-07-09
Updates release workflows to use nuget trusted publishing

Adam-it:
## 🎯 Aim

The aim is to refactor the current workflows that interact with nuget.org to use short-lived tokens based on trusted publishing policies


gautamdsheth:
LGTM, once we have the policy in place in Nuget, we can merge it !

Adam-it:
@jansenbe I will assign you to this PR as I am unable to push it forward without your actions in the NuGet.org service

PR #1802, opened 2026-07-15
#1799 - CAML queries with Joins/ProjectedFields do not return projected fields - add CSOM-based GetItems support. Closes #1799

### LINKED ISSUE #1799: Feature: CAML queries with Joins/ProjectedFields do not return projected fields - add CSOM-based GetItems support
## Category

- [x] Feature request

## Describe the feature

When executing a CAML query containing `<Joins>` and `<ProjectedFields>` through `IList.LoadItemsByCamlQueryAsync` (or its sync/batch variants), the fields projected from the joined list are not present in the returned items. The underlying REST `GetItems` endpoint does not serialize projected fields, so there is currently no way in the PnP Core SDK to retrieve joined-list fields from a CAML join query.

Example query where `ProjectedText` is never returned by the REST based methods:

```xml
<View>
  <ViewFields>
    <FieldRef Name='Title'/><FieldRef Name='JoinLookup'/><FieldRef Name='ProjectedText'/>
  </ViewFields>
  <Joins>
    <Join Type='LEFT' ListAlias='JoinedList'>
      <Eq>
        <FieldRef Name='JoinLookup' RefType='Id'/>
        <FieldRef List='JoinedList' Name='Id'/>
      </Eq>
    </Join>
  </Joins>
  <ProjectedFields>
    <Field Name='ProjectedText' Type='Lookup' List='JoinedList' ShowField='SourceText'/>
  </ProjectedFields>
</View>
```

The CSOM List.GetItems(CamlQuery) method does return these fields, since the CSOM response serializes the full field dictionary of each list item.

Describe the solution you'd like

Add CSOM-backed methods to IList:

```csharp
Task<ICamlQueryCsomResult> LoadItemsByCamlQueryViaCsomAsync(CamlQueryOptions queryOptions);
ICamlQueryCsomResult LoadItemsByCamlQueryViaCsom(CamlQueryOptions queryOptions);
```

- Executes the CAML query via CSOM List.GetItems(CamlQuery), so fields from Joins/ProjectedFields are returned and typed as IFieldLookupValue.
- Items are materialized into the list's Items collection, consistent with the existing LoadItemsByCamlQuery methods.
- The result exposes the loaded items plus PagingInfo (taken from the CSOM ListItemCollectionPosition), which feeds straight back into CamlQueryOptions.PagingInfo to load the next page - no need to construct paging info manually.
- Batch variants are intentionally omitted: the CSOM result is materialized immediately after execution, which does not fit the deferred batch model.

Additional context

I have a working implementation ready, including:

- CSOM request unit tests validating the generated CSOM XML and response parsing
- An integration test (join + projected field + paging) with recorded mock data, so it runs offline

I'd like to contribute this as a PR once the feature scope is agreed.

jeppesc11:
I added LoadItemsByCamlQueryViaCsomAsync / LoadItemsByCamlQueryViaCsom to IList. 

These execute the CAML query via the CSOM List.GetItems(CamlQuery) method instead of the REST GetItems endpoint, so fields projected from joined lists (CAML Joins/ProjectedFields) are now returned as well, typed as IFieldLookupValue. 

The loaded items are merged into the list's Items collection like the existing LoadItemsByCamlQuery methods, and the result also returns the PagingInfo from the CSOM ListItemCollectionPosition, which you can set on CamlQueryOptions.PagingInfo to load the next page without constructing the paging string manually.

## 🔗 Related Issue

Closes #1799


Adam-it:
Thanks @jeppesc11 for opening the PR. We will try to look into it ASAP

src/sdk/PnP.Core/Model/SharePoint/Core/Internal/CsomListItemsHandler.cs
```diff
@@ -0,0 +1,218 @@
...
+                    return null;
+                case JsonValueKind.True:
+                case JsonValueKind.False:
+                    return value.GetBoolean();
+                case JsonValueKind.Number:
+                    if (value.TryGetInt32(out int intValue))
+                    {
+                        return intValue;
+                    }
+                    return value.GetDouble();
```
Adam-it: I was wondering if doing something similar to what we already do here https://github.com/pnp/pnpcore/blob/6650344b52b09a7372f06d9f584a7e7f4c75d320/src/sdk/PnP.Core/Model/SharePoint/Core/Internal/ListDataAsStreamHandler.cs#L589-L615

would make more sense. The main reason is that a Number field holding 5 comes back as int here but as double via LoadItemsByCamlQueryAsync/GetListDataAsStreamAsync, so existing (double)item["MyNumber"] code throws InvalidCastException depending on which loader was used.

src/sdk/PnP.Core/Model/SharePoint/Core/Internal/CsomListItemsHandler.cs
```diff
@@ -0,0 +1,218 @@
...
+                                }
+                            }
+                            else
+                            {
+                                stringValues ??= new List<string>();
+                                stringValues.Add(element.ValueKind == JsonValueKind.String ? element.GetString() : element.GetRawText());
+                            }
+                        }
+
+                        return (object)fieldValueCollection ?? stringValues ?? new List<string>();
```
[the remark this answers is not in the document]
Adam-it: This is a valid comment IMO 👍

src/sdk/PnP.Core/Model/SharePoint/Core/Internal/CsomListItemsHandler.cs
```diff
@@ -0,0 +1,218 @@
...
+                objectType = objectTypeElement.GetString();
+            }
+
+            switch (objectType)
+            {
+                case "SP.FieldLookupValue": return new FieldLookupValue { Field = field };
+                case "SP.FieldUserValue": return new FieldUserValue { Field = field };
+                case "SP.FieldUrlValue": return new FieldUrlValue { Field = field };
+                case "SP.Taxonomy.TaxonomyFieldValue": return new FieldTaxonomyValue { Field = field };
+                default: return null;
```
Adam-it: I wonder if we shouldn't add support for other fields like we already do for the rest handler in order to keep it consistent 
https://github.com/pnp/pnpcore/blob/6650344b52b09a7372f06d9f584a7e7f4c75d320/src/sdk/PnP.Core/Model/SharePoint/Core/Internal/ListDataAsStreamHandler.cs#L514-L522

src/sdk/PnP.Core/Model/SharePoint/Core/Internal/List.cs
```diff
@@ -457,6 +458,35 @@ private async Task<ApiCall> BuildGetItemsByCamlQueryApiCall(CamlQueryOptions que
...
+            if (queryOptions == null)
+            {
+                throw new ArgumentNullException(nameof(queryOptions));
+            }
+
+            var request = new GetItemsByCamlQueryRequest(PnPContext.Site.Id.ToString(), PnPContext.Web.Id.ToString(), Id.ToString(), queryOptions);
+
+            var apiCall = new ApiCall(new List<Services.Core.CSOM.Requests.IRequest<object>>() { request })
+            {
+                Commit = true,
```
Adam-it: Wouldn't it be better if we set `commit` to `true`  only on write calls 🤔. I believe this may silently discarding any pending (unsaved) property changes on that list

src/sdk/PnP.Core/Model/SharePoint/Core/Internal/CsomListItemsHandler.cs
```diff
@@ -0,0 +1,218 @@
...
+
+                    if (!overflowDictionary.ContainsKey(property.Name))
+                    {
+                        overflowDictionary.SystemAdd(property.Name, fieldValue);
+                    }
+                    else
+                    {
+                        overflowDictionary.SystemUpdate(property.Name, fieldValue);
+                    }
+                }
```
Adam-it, line 91: aren't we missing out a few properties here? If we check the `ListDataAsStreamHandler`
https://github.com/pnp/pnpcore/blob/6650344b52b09a7372f06d9f584a7e7f4c75d320/src/sdk/PnP.Core/Model/SharePoint/Core/Internal/ListDataAsStreamHandler.cs#L193-L245

We also map stuff like `FSObjType` or `UniqueId` which are as well present in the CamlQueryViaCsom results.
Values and the model property backing store are two different dictionaries. `ListItem` derives from `ExpandoBaseDataModel<IListItem>`, where `Values` is itself stored as one entry `(current["Values"])` of the `TransientObject` backing store. So writing `Values["UniqueId"]` does not make `HasValue("UniqueId")` `true`.

Adam-it (changes requested):
@jeppesc11 amaizing work 👏. 
Checked locally ✅
I only left a few concerns about some edge cases I noticed that I think it's worth recheking before we merge.
I would also suggest we extend the docs https://github.com/pnp/pnpcore/blob/dev/docs/using-the-sdk/listitems-intro.md providing new examples how to use the added methods in this PR

src/sdk/PnP.Core/Model/SharePoint/Core/Internal/CsomListItemsHandler.cs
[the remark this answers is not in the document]
jeppesc11: Absolutely! 😊

src/sdk/PnP.Core/Model/SharePoint/Core/Internal/CsomListItemsHandler.cs
jeppesc11: You're absolutely right - that was an oversight on my part. 😅

src/sdk/PnP.Core/Model/SharePoint/Core/Internal/List.cs
jeppesc11: Agreed. Looking at it again, mine is the only read call that sets commit to true - everywhere else it's false or omitted. I'll drop it.

src/sdk/PnP.Core/Model/SharePoint/Core/Internal/CsomListItemsHandler.cs
jeppesc11: Agreed - I got a bit tunnel-visioned on my own use case...

src/sdk/PnP.Core/Model/SharePoint/Core/Internal/CsomListItemsHandler.cs
jeppesc11: Good catch, agreed.

jeppesc11:
Thanks for the thorough review @Adam-it 🙏 I've pushed a commit addressing the feedback.

Adam-it (approved):
LGTM 🔥

PR #1819, opened 2026-07-27
Expose list item validation error details

### LINKED ISSUE #1764: Update the `HandleValidateAddAndUpdateResponse` in the `ListItem` to pass on the error code and error message if present
That is some good question and feedback. 
I think the problem is that here 
https://github.com/pnp/pnpcore/blob/07bd47676e8c60e5ccf8c8e89422efa6d7a125d9/src/sdk/PnP.Core/Model/SharePoint/Core/Internal/ListItem.cs#L190-L217
We just read the `HasException` and handle the error but we do not parse and pass one the error code and error message. 
What we could do is:
- Read ErrorCode (int) and ErrorMessage (string) from the field JsonElement
- Construct a SharePointRestError with those details (or pass them into the message)

based on: #1760

Guflly:
## Aim

Expose the SharePoint error code and message returned by list item validation.

## What was changed

`SharePointRestError` now receives those details for both thrown exceptions and non-throwing batch results. Existing list item tests cover both paths.

## Related issue

Closes #1764

### DIFF
--- a/src/sdk/PnP.Core.Test/SharePoint/ListItemTests.cs
+++ b/src/sdk/PnP.Core.Test/SharePoint/ListItemTests.cs
@@ -303,10 +303,15 @@ namespace PnP.Core.Test.SharePoint
                     await list.Items.AddBatchAsync(propItem);
                 }
 
-                await Assert.ThrowsExceptionAsync<SharePointRestServiceException>(async () =>
+                var exception = await Assert.ThrowsExceptionAsync<SharePointRestServiceException>(async () =>
                 {
                     await context.ExecuteAsync();
                 });
+
+                var error = exception.Error as SharePointRestError;
+                Assert.IsNotNull(error);
+                Assert.AreEqual(0, error.ServerErrorCode);
+                Assert.AreEqual("abc", error.Message);
 
                 await list.DeleteAsync();
             }
@@ -354,6 +359,10 @@ namespace PnP.Core.Test.SharePoint
                     var errors = await context.ExecuteAsync(false);
 
                     Assert.IsTrue(errors.Count == 2);
+                    var error = errors[0].Error as SharePointRestError;
+                    Assert.IsNotNull(error);
+                    Assert.AreEqual(0, error.ServerErrorCode);
+                    Assert.AreEqual("abc", error.Message);
                 }
                 finally
                 {
--- a/src/sdk/PnP.Core.Test/SharePoint/ListItemTests.cs
+++ b/src/sdk/PnP.Core.Test/SharePoint/ListItemTests.cs
@@ -416,6 +416,10 @@ namespace PnP.Core.Test.SharePoint
                     var errors = await context.ExecuteAsync(batch, false);
 
                     Assert.IsTrue(errors.Count == 2);
+                    var error = errors[0].Error as SharePointRestError;
+                    Assert.IsNotNull(error);
+                    Assert.AreEqual(0, error.ServerErrorCode);
+                    Assert.AreEqual("abc", error.Message);
                 }
                 finally
                 {
--- a/src/sdk/PnP.Core/Model/SharePoint/Core/Internal/ListItem.cs
+++ b/src/sdk/PnP.Core/Model/SharePoint/Core/Internal/ListItem.cs
@@ -196,8 +196,13 @@ namespace PnP.Core.Model.SharePoint
                 var errorMessage = field.TryGetProperty("ErrorMessage", out JsonElement errorMessageProperty)
                     && errorMessageProperty.ValueKind == JsonValueKind.String
                     ? errorMessageProperty.GetString()
-                    : string.Format(PnPCoreResources.Exception_ListItemAdd_WrongInternalFieldName, fieldName);
+                    : null;
+                if (string.IsNullOrEmpty(errorMessage))
+                {
+                    errorMessage = string.Format(PnPCoreResources.Exception_ListItemAdd_WrongInternalFieldName, fieldName);
+                }
                 var errorCode = field.TryGetProperty("ErrorCode", out JsonElement errorCodeProperty)
+                    && errorCodeProperty.ValueKind == JsonValueKind.Number
                     && errorCodeProperty.TryGetInt64(out long parsedErrorCode) ? parsedErrorCode : 0;
                 var error = new SharePointRestError(ErrorType.SharePointRestServiceError, (int)System.Net.HttpStatusCode.OK, errorCode, errorMessage);
 
@@ -218,10 +223,7 @@ namespace PnP.Core.Model.SharePoint
 
                 if (!handled)
                 {
-                    throw new SharePointRestServiceException(errorMessage)
-                    {
-                        Error = error
-                    };
+                    throw new SharePointRestServiceException(errorMessage, error);
                 }
             }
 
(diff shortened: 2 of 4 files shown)


Adam-it:
@Guflly we do appreciate your efforts, but before we start please do follow the contributing guide best practices for our repo
https://pnp.github.io/pnpcore/index.html#best-practices-when-contributing
First, please leave a comment in the related issue so that you may get assigned to it and only then lets get back to this PR

Guflly:
Done — I commented on #1764 and asked to be assigned.

src/sdk/PnP.Core/Model/SharePoint/Core/Internal/ListItem.cs
```diff
@@ -193,6 +193,14 @@ private void HandleValidateAddAndUpdateResponse(JsonElement field, JsonElement v
             // not be added which is indicated via the HasException property on the field. If so then throw an error.
             if (field.TryGetProperty("HasException", out JsonElement hasExceptionProperty) && hasExceptionProperty.GetBoolean() == true)
             {
+                var errorMessage = field.TryGetProperty("ErrorMessage", out JsonElement errorMessageProperty)
+                    && errorMessageProperty.ValueKind == JsonValueKind.String
+                    ? errorMessageProperty.GetString()
+                    : string.Format(PnPCoreResources.Exception_ListItemAdd_WrongInternalFieldName, fieldName);
+                var errorCode = field.TryGetProperty("ErrorCode", out JsonElement errorCodeProperty)
+                    && errorCodeProperty.TryGetInt64(out long parsedErrorCode) ? parsedErrorCode : 0;
```
Adam-it, line 206: lets add a ValueKind guard.
`JsonElement.TryGetInt64` throws `InvalidOperationException` when `ValueKind` is not Number — it does not return `false`. 
Lets try something like `jsonElement.ValueKind == JsonValueKind.Number && jsonElement.TryGetInt64(...)` here. We do similar check in other places in our codebase already

src/sdk/PnP.Core/Model/SharePoint/Core/Internal/ListItem.cs
```diff
@@ -203,14 +211,17 @@ private void HandleValidateAddAndUpdateResponse(JsonElement field, JsonElement v
...
                     }
                 }
 
                 if (!handled)
                 {
-                    throw new SharePointRestServiceException(string.Format(PnPCoreResources.Exception_ListItemAdd_WrongInternalFieldName, fieldName));
+                    throw new SharePointRestServiceException(errorMessage)
+                    {
+                        Error = error
+                    };
```
Adam-it: Let's not assign Error via object initializer from the calling code
Instead, let's add something like internal SharePointRestServiceException(string message, SharePointRestError error) (or an overload taking type/code/errorCode/message) to SharePointRestServiceException.cs and keep the Error assignment there.

src/sdk/PnP.Core/Model/SharePoint/Core/Internal/ListItem.cs
```diff
@@ -193,6 +193,14 @@ private void HandleValidateAddAndUpdateResponse(JsonElement field, JsonElement v
             // not be added which is indicated via the HasException property on the field. If so then throw an error.
             if (field.TryGetProperty("HasException", out JsonElement hasExceptionProperty) && hasExceptionProperty.GetBoolean() == true)
             {
+                var errorMessage = field.TryGetProperty("ErrorMessage", out JsonElement errorMessageProperty)
+                    && errorMessageProperty.ValueKind == JsonValueKind.String
```
Adam-it, line 197: I think the  `"ErrorMessage": ""` yields an empty exception message instead of the fallback. lets add a `string.IsNullOrEmpty` check on the extracted value as well

src/sdk/PnP.Core.Test/SharePoint/ListItemTests.cs
Adam-it, line 375: Shouldn't we modify this test as well?

Adam-it (changes requested):
@Guflly Awesome work 👏
I found some tiny details we could refine before we make this PR complete and process it further. Please take a look at the few of my comments 👍
You Rock 🤩

src/sdk/PnP.Core/Model/SharePoint/Core/Internal/ListItem.cs
Guflly: Added the empty-message fallback.

src/sdk/PnP.Core/Model/SharePoint/Core/Internal/ListItem.cs
Guflly: Moved the Error assignment into the new exception constructor.

src/sdk/PnP.Core.Test/SharePoint/ListItemTests.cs
Guflly: Updated the named batch test too.

src/sdk/PnP.Core/Model/SharePoint/Core/Internal/ListItem.cs
Guflly: Added the ValueKind guard.

src/sdk/PnP.Core/Model/SharePoint/Core/Internal/ListItem.cs
```diff
@@ -203,14 +216,14 @@ private void HandleValidateAddAndUpdateResponse(JsonElement field, JsonElement v
...
-                                                 new SharePointRestError(ErrorType.SharePointRestServiceError, (int)System.Net.HttpStatusCode.OK, string.Format(PnPCoreResources.Exception_ListItemAdd_WrongInternalFieldName, fieldName)));
+                                                 error);
                         handled = true;
                     }
                 }
 
                 if (!handled)
                 {
-                    throw new SharePointRestServiceException(string.Format(PnPCoreResources.Exception_ListItemAdd_WrongInternalFieldName, fieldName));
+                    throw new SharePointRestServiceException(errorMessage, error);
```
Adam-it, line 226: aaj I just noticed that we will lose the previous message that was Exception_ListItemAdd_WrongInternalFieldName → "There was an exception while writing field {0}. Verify you're using the correct InternalName value for the field you want to write to." Now, whenever SharePoint returns a non-empty ErrorMessage, that resource string is dropped entirely and replaced by the raw server text 🤔. That means if SharePoint returns "ErrorMessage":"abc", then the thrown exception message becomes just "abc", with no indication that MetaInfo was the offending field or that the InternalName is likely wrong. 
I would keep the resource string as the message and append/attach the server detail, e.g. message = string.Format(PnPCoreResources.Exception_ListItemAdd_WrongInternalFieldName, fieldName) plus the server text (or surface the raw server message via AddAdditionalData, as CsomError does for ErrorValue/ErrorTypeName), while ServerErrorCode carries the code.

Adam-it (changes requested):
@Guflly I think we are on the right track. When testing around I noticed we are losing previous error description which was really helpful. I would refactor a bit to try to aling with what we do in CSOMError handling adding the SP error message and value as additional data to the exception. it would be best to try to reproduce the same behavior as well. 
Sorry for the late observation and feed on this 👍

Adam-it:
@Guflly just a kind reminder and recheck if you managed to check my previous comment in the PR?

PR #1839, opened 2026-08-12
Scope ISharePointUser.Groups to the groups the user belongs to

### LINKED ISSUE #1630: Getting the groups of site user returns all the site groups instead of only the groups he's a member of
## Category

- [x] Bug

## Describe the bug

As the title says, I am trying to get all the groups that a user is a member of on a given SPO site, but instead of only getting "his" groups, I get all the site groups, even if the user is not a member of said groups.

## Steps to reproduce

```
using var siteContext = await CreateContextAsync(siteRelativeUrl);

var lowerCaseUpn = userPrincipalName.ToLower();

var user = await siteContext.Web.SiteUsers
    .QueryProperties(p => p.UserPrincipalName)
    .FirstOrDefaultAsync(u => u.UserPrincipalName == lowerCaseUpn)
    ?? throw new SpServiceException("The user you're trying to query was not found.");

// This returns all the site groups, whether the user is a member of them or not
var userGroups = await user.Groups.ToListAsync();
```

## Expected behavior

I should only get the groups he's a member of (directly OR in a transitive way (if he's inside an Entra group that is a member of the group).

svermaak:
Fixes #1630

### What is happening

`ISharePointUser.Groups` returns every group in the site, whether or not the user is a member of any of them.

`SharePointGroup` carried a single `[SharePointType("SP.Group", ...)]` mapping with `LinqGet = "_api/Web/SiteGroups"`. That mapping is used no matter which model the group collection hangs off, so `user.Groups` issued the site wide request and the user was never part of the query.

### Repro

Test site with 5 site groups. `shaunvermaak@ittelligencedev.onmicrosoft.com` was added to `PnPCore1630 Member Group` and to nothing else, then `user.Groups.ToListAsync()` was called.

```csharp
var user = await context.Web.EnsureUserAsync(upn);
var userGroups = await user.Groups.ToListAsync();
```

Before - all 5 site groups come back, including one the user is explicitly not a member of:

before

After - only the group the user actually belongs to:

after

### The fix

`SP.Group` is now mapped per parent type, with the `SharePointUser` mapping pointing at the user scoped endpoint:

```csharp
[SharePointType("SP.Group", Target = typeof(Web), ... LinqGet = "_api/Web/SiteGroups", ...)]
[SharePointType("SP.Group", Target = typeof(Site), ... LinqGet = "_api/Web/SiteGroups", ...)]
[SharePointType("SP.Group", Target = typeof(SharePointUser), ... LinqGet = "_api/Web/GetUserById({Parent.Id})/Groups", ...)]
```

The explicit `Web` and `Site` mappings are required, not cosmetic. `EntityInfo` only resolves a mapping by exact parent match once a model carries more than one `SharePointType` attribute, so leaving the original attribute untargeted would have broken `Web.SiteGroups`, `Web.Associated*Group` and `Site.HubSiteSynchronizableVisitorGroup`. This mirrors how `SP.User`, `SP.Field` and `SP.ContentType` are already mapped.

`Uri`, `Get`, `Update` and `Delete` stay on `_api/Web/sitegroups/...` for every parent, so only the enumeration scope changes.

### Tests

Added `SharePointGroupTests.SharePointGroupEndpointDependsOnParent`, an offline test asserting the resolved endpoint for each of the three parent types. A mocked test would not have caught this: mock responses are replayed by request sequence, so the wrong URL still replays the right payload.

The new test fails on `dev` and passes with this change.

Full offline test suite before and after this change: 8 identical pre-existing failures (timezone, access token, clone and page publish tests), no new failures.

No CHANGELOG entry included.

### DIFF
--- a/src/sdk/PnP.Core.Test/Security/SharePointGroupTests.cs
+++ b/src/sdk/PnP.Core.Test/Security/SharePointGroupTests.cs
@@ -1,4 +1,5 @@
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using PnP.Core.Model;
 using PnP.Core.Model.Security;
 using PnP.Core.QueryModel;
 using PnP.Core.Test.Utilities;
@@ -475,6 +476,25 @@ namespace PnP.Core.Test.Security
                         await siteGroup.DeleteAsync();
                     }
                 }
+            }
+        }
+
+        [TestMethod]
+        public async Task SharePointGroupEndpointDependsOnParent()
+        {
+            // Offline test: mocked responses are replayed per request sequence, so only asserting on the
+            // generated endpoint catches a group collection that is loaded from the wrong scope
+            using (var context = await TestCommon.Instance.GetContextWithoutInitializationAsync(TestCommon.TestSite))
+            {
+                var user = new SharePointUser { PnPContext = context, Parent = context.Web, Id = 12 };
+
+                var fromWeb = EntityManager.GetClassInfo<ISharePointGroup>(typeof(SharePointGroup), null, parent: context.Web as IDataModelParent);
+                var fromSite = EntityManager.GetClassInfo<ISharePointGroup>(typeof(SharePointGroup), null, parent: context.Site as IDataModelParent);
+                var fromUser = EntityManager.GetClassInfo<ISharePointGroup>(typeof(SharePointGroup), null, parent: user);
+
+                Assert.AreEqual("_api/Web/SiteGroups", fromWeb.SharePointLinqGet);
+                Assert.AreEqual("_api/Web/SiteGroups", fromSite.SharePointLinqGet);
+                Assert.AreEqual("_api/Web/GetUserById({Parent.Id})/Groups", fromUser.SharePointLinqGet);
             }
         }
     }
--- a/src/sdk/PnP.Core/Model/Security/Internal/SharePointGroup.cs
+++ b/src/sdk/PnP.Core/Model/Security/Internal/SharePointGroup.cs
@@ -1,4 +1,5 @@
-﻿using PnP.Core.QueryModel;
+﻿using PnP.Core.Model.SharePoint;
+using PnP.Core.QueryModel;
 using PnP.Core.Services;
 using PnP.Core.Utilities;
 using System;
@@ -9,7 +10,9 @@ using System.Threading.Tasks;
 
 namespace PnP.Core.Model.Security
 {
-    [SharePointType("SP.Group", Uri = "_api/Web/sitegroups/getbyid({Id})", LinqGet = "_api/Web/SiteGroups", Delete = "_api/Web/SiteGroups/RemoveById({Id})")]
+    [SharePointType("SP.Group", Target = typeof(Web), Uri = "_api/Web/sitegroups/getbyid({Id})", LinqGet = "_api/Web/SiteGroups", Delete = "_api/Web/SiteGroups/RemoveById({Id})")]
+    [SharePointType("SP.Group", Target = typeof(Site), Uri = "_api/Web/sitegroups/getbyid({Id})", LinqGet = "_api/Web/SiteGroups", Delete = "_api/Web/SiteGroups/RemoveById({Id})")]
+    [SharePointType("SP.Group", Target = typeof(SharePointUser), Uri = "_api/Web/sitegroups/getbyid({Id})", LinqGet = "_api/Web/GetUserById({Parent.Id})/Groups", Delete = "_api/Web/SiteGroups/RemoveById({Id})")]
     internal sealed class SharePointGroup : BaseDataModel<ISharePointGroup>, ISharePointGroup
     {
         #region Construction


Adam-it:
@svermaak we are happy you are eager to help us out but please do follow our [contribution guide](https://github.com/pnp/pnpcore/blob/dev/CONTRIBUTING.md) and [contributing best practices](https://pnp.github.io/pnpcore/index.html#best-practices-when-contributing).

The issue for which you opened a PR was not ready, it does not have the 'help wanted' label which means we did not agree on the fix direction and were not expecting anyone working on it.
Also you did not leave any comment in the issue do first get assigned before opening the PR. Please first get assigned to the PR and lets move the discussion first there to agree on the fix before we process this PR further

svermaak:
image

Adam-it:
@svermaak thanks for pointing out the failing link in the contributing guide. I will refresh and sort it out in a few min and make the guide cleaner. Nevertheless, the links I provided in my previous comment work and explain the path that should be followed, which will help us keep this repo maintainable instead of having a flood of open and stale PRs that could never get merged 👍

Adam-it:
Same exercise as on #1840: revisiting this one against #1630 so the analysis is recorded before the issue gets picked up again. Short version: I think this is on the right track and the root cause is correctly identified, with one scope caveat and the same reservation about the test.

### The diagnosis holds up

`EntityInfo.SharePoint` (`src/sdk/PnP.Core/Model/Base/EntityInfo.cs:234-253`) returns the single mapping when there is exactly one `SharePointType` attribute, regardless of parent. With `SP.Group` carrying only

```csharp
[SharePointType("SP.Group", Uri = "...", LinqGet = "_api/Web/SiteGroups", Delete = "...")]
```

`QueryClient.BuildGetAPICallRestAsync` resolves `SharePointLinqGet` to `_api/Web/SiteGroups` for every parent, so `user.Groups.ToListAsync()` really does issue the site wide query with the user nowhere in it. That matches the report exactly.

The claim in the PR description that the `Web` and `Site` mappings are required rather than cosmetic is also correct, and worth repeating because it is the non obvious part: once `SharePointTargets.Count > 1`, resolution falls to `SharePointTargets.FirstOrDefault(p => p.Target == Target)` and returns **null** on no match. `EntityManager` sets `entityInfo.Target = parent.GetType()` (`EntityManager.cs:415-418`), hopping over the collection first, so the match is on the concrete parent type. Leaving the original attribute untargeted would have silently null'd the mapping for `Web.SiteGroups`, `Web.Associated*Group` and `Site.HubSiteSynchronizableVisitorGroup`.

I checked the parent surface and those three are the complete set of parents for `ISharePointGroup` / `ISharePointGroupCollection` in the SDK: `Web.SiteGroups` and `Web.Associated{Member,Owner,Visitor}Group` (`Web.cs:312,320-324`), `Site.HubSiteSynchronizableVisitorGroup` (`Site.cs:142`), `SharePointUser.Groups` (`SharePointUser.cs:76`). Nothing in `PnP.Core.Admin` and no other construction site. So the three attributes cover it, and per parent mapping is the established convention here anyway, `SP.User` right next door does exactly this (`SharePointUser.cs:12-13`).

Keeping `Uri` / `Get` / `Delete` on `_api/Web/sitegroups/...` for all three is right: reading or deleting an individual group is site scoped, only the enumeration scope was wrong.

### The scope caveat

The issue asks for the groups the user belongs to "directly OR in a transitive way (if he's inside an Entra group that is a member of the group)". `_api/Web/GetUserById(id)/Groups` returns direct SharePoint group membership only, it will not surface a group the user reaches through an Entra security group that is itself a member. So this fixes the reported bug (all site groups coming back) but only the first half of the stated expectation. That is probably the right call, since transitive expansion is not something that endpoint can do and would need Graph, but it should be an explicit decision rather than an implicit one, and the XML doc on `ISharePointUser.Groups` (`ISharePointUser.cs:55-57`) would be the place to say "direct membership".

### The test, same reservation as on #1840

`SharePointGroupEndpointDependsOnParent` asserts `SharePointLinqGet` string values for the three parents, which is close to reading the attributes back. The point about mocked tests being unable to catch this is accurate, `TestManager` keys mocks by test name plus request order (`{TestName}-{TestId}-{orderPrefix}.response.json`), so a wrong url replays the right payload. But the assertion here stops at an **unresolved** `{Parent.Id}` token, and that token is the part carrying real risk: `TokenHandler` resolves `{Parent.Id}` from the parent's `Metadata[MetaDataRestId]`, not from the `Id` property (`TokenHandler.cs:67-88`). If a `SharePointUser` reaches `user.Groups` without that metadata populated, the url ships with the literal token in it. The test as written cannot see that.

`PnP.Core.Test/Base/QueryTests.cs` already has the better pattern for this: `BuildModel` plus `GetAPICallTestAsync` call `QueryClient.BuildGetAPICallAsync` and assert the generated request url after token resolution. Asserting the fully resolved `_api/Web/GetUserById(12)/Groups` for a user parented collection would cover both the mapping and the token, and would fail if either regressed.

### What still needs checking

- A user reached through paths other than the repro: `web.CurrentUser`, `SiteUsers.FirstOrDefaultAsync(...)` as in the issue, `EnsureUserAsync`, and a user obtained from `group.Users` (that one resolves the `Target = typeof(SharePointGroup)` `SP.User` mapping, so it is worth confirming `{Parent.Id}` still resolves off it).
- `web.CurrentUser.QueryProperties(p => p.Groups)` is already covered by `UserTests.cs:59` and goes through `$expand` rather than `LinqGet`, so it should be unaffected, but it is a cheap regression check.
- Paging behaviour on the user scoped endpoint for a user in many groups.

Overall the direction looks correct to me and consistent with how the model already handles multi parent types. @svermaak, as on the other one, sorry about how the earlier exchange landed. If you want to pick this up once the issue is labelled `help wanted`, this is a good base.

PR #1840, opened 2026-08-12
Read a sub folder from its own endpoint instead of its parent's

### LINKED ISSUE #1586: Enumerating child objects of document library folder shows parent objects with AsRequested
## Category

- [x] Bug

## Describe the bug

If I have a folder inside a document library in SharePoint, and enumerate the child objects (files or folders) of that folder, using `.AsRequested()`, it returns objects of the root folder, not the specified folder.

I have a SharePoint document library named "SDK Test" with the following structure:
```
Top-level file.txt
Top-level Folder\
    Nested file.txt
    Nested Folder\
```
image

If I enumerate the Files or Folders on the document library (using the RootFolder as shown in [documentation](https://pnp.github.io/pnpcore/using-the-sdk/files-intro.html#enumerating-files)), it works correctly and returns the same results whether or not I include the `.AsRequested()`.

However if I get a reference to the [document folder](https://pnp.github.io/pnpcore/using-the-sdk/folders-intro.html#getting-folders) ("Top-level Folder") and enumerate its children, I get different results if I use `.AsRequested()`: it shows the top-level files/folders.

The [comments](https://pnp.github.io/pnpcore/using-the-sdk/basics-getdata.html#i-want-to-load-data-into-the-pnpcontext) on `AsRequested()` indicate it simply uses previously-loaded data to prevent another round-trip to SharePoint:
```cs
// We're using AsRequested() to query the already loaded domain models, if not a new query would 
// issued to load the lists
```
As such I can see that `.AsRequested` may have the capability to return cached data _if called first_, but:
1. I'm calling `.Files` directly before calling `.AsRequested()`
2. It's returning data from the parent document library, not the specified folder.

Also the order of calling `.Files` or `.Files.AsRequested()` doesn't matter.

## Steps to reproduce

This code can be used to replicate the issue:
```cs
public async Task TestSdk()
{
    using PnPContext context = _pnpContextFactory.Create("Default");

    IFolder libraryFolder = (await context.Web.Lists.GetByTitleAsync("SDK Test", p => p.RootFolder)).RootFolder;
    await libraryFolder.LoadAsync(p => p.Folders, p => p.Files);
    Console.WriteLine("Library Folder");
    Console.WriteLine("--------------");
    PrintDebug(libraryFolder);

    Console.WriteLine();

    IFolder topLevelFolder = libraryFolder.Folders.FirstOrDefault(p => p.Name == "Top-level Folder")!;
    await topLevelFolder.LoadAsync(p => p.Folders, p => p.Files);
    Console.WriteLine("Top-level Folder");
    Console.WriteLine("----------------");
    PrintDebug(topLevelFolder);
}

public void PrintDebug(IFolder folder)
{
    foreach (var f in folder.Folders)
    {
        Console.WriteLine($"Folders: {f.Name}");
    }
    foreach (var f in folder.Folders.AsRequested())
    {
        Console.WriteLine($"Folders AsRequested: {f.Name}");
    }
    foreach (var f in folder.Files)
    {
        Console.WriteLine($"Files: {f.Name}");
    }
    foreach (var f in folder.Files.AsRequested())
    {
        Console.WriteLine($"Files AsRequested: {f.Name}");
    }
}
```

And this is the output:
```
Library Folder
--------------
Folders: Top-level Folder
Folders: Forms
Folders AsRequested: Top-level Folder
Folders AsRequested: Forms
Files: Top-level file.txt
Files AsRequested: Top-level file.txt

Top-level Folder
----------------
Folders: Nested Folder
Folders AsRequested: Top-level Folder     <-- Wrong
Folders AsRequested: Forms                <-- Wrong
Files: Nested file.txt
Files AsRequested: Top-level file.txt     <-- Wrong
```

## Expected behavior

`AsRequested()` can return cached or old for the specified folder ("Top-level folder"), but it should not be returning data for the library ("SDK Test").

## Environment details (*development & target environment*)

- **SDK version**: 1.14.0
- **OS**: Docker devcontainer mcr.microsoft.com/devcontainers/dotnet:8.0 based on Debian 12.7
- **SDK used in**: Class library accessed by a Console App.
- **Framework**: .NET Core 8.0.403
- **Browser(s)**: N/A
- **Tooling**: VS Code
- **Additional details**: N/A

svermaak:
Fixes #1586

### What is happening

The `SP.Folder` mapping for a folder whose parent is a folder overrides `Get`:

```csharp
[SharePointType("SP.Folder", Target = typeof(Folder), Uri = "_api/Web/getFolderById('{Id}')", Get = "_api/Web/getFolderById('{Parent.Id}')", ...)]
```

`Get` is what every non LINQ read of the model resolves to, so reading a sub folder actually requested its **parent's** url. `await subFolder.LoadAsync(p => p.Folders, p => p.Files)` therefore returned the parent folder's children, and those are the items `AsRequested()` hands back. The `Folders` and `Files` queryables were unaffected because they resolve `LinqGet`, which is correct - that is exactly the discrepancy reported in the issue.

It is not limited to collections. Loading any single property on a sub folder returns the parent's value, as the second block in the screenshots shows.

### Repro

Library `PnPCore1586` with the structure from the issue:

```
Top-level file.txt
Top-level Folder/
    Nested file.txt
    Nested Folder/
```

```csharp
await root.LoadAsync(p => p.Folders, p => p.Files);
IFolder top = root.Folders.AsRequested().First(f => f.Name == "Top-level Folder");
await top.LoadAsync(p => p.Folders, p => p.Files);
// top.Folders.AsRequested() / top.Files.AsRequested()
```

Before - the library root's children come back, and `ServerRelativeUrl` on a freshly created sub folder reads back the parent's url:

before

After:

after

### The fix

`Get` is dropped from that mapping so it falls back to `Uri`, `_api/Web/getFolderById('{Id}')`.

`AddApiCallHandler` was the one consumer that depended on `Get` pointing at the parent, since a folder is created through its container: `{container}/Folders/AddUsingPath(...)`. It now builds that parent url itself when the target is a folder, so adding folders is unchanged. The screenshots include an add plus read-back of the new folder to confirm both halves.

`SP.File` with `Target = typeof(Folder)` already had no `Get` override, so files were only affected through the expand on the folder read.

### Tests

Added `FoldersTests.SubFolderIsReadFromItsOwnEndpoint`, an offline test pinning `Get` and `LinqGet` for the folder-in-folder mapping. A mocked test cannot catch this: mock responses are replayed by request sequence, so the wrong url still replays the right payload.

The new test fails on `dev` and passes with this change.

Full offline test suite: 8 pre-existing failures (timezone, access token, clone and page publish tests) before and after, no new failures.

No CHANGELOG entry included.

### DIFF
--- a/src/sdk/PnP.Core.Test/SharePoint/FoldersTests.cs
+++ b/src/sdk/PnP.Core.Test/SharePoint/FoldersTests.cs
@@ -1486,5 +1486,25 @@ namespace PnP.Core.Test.SharePoint
         }
 
         #endregion
+
+        #region Endpoint mapping
+
+        [TestMethod]
+        public async Task SubFolderIsReadFromItsOwnEndpoint()
+        {
+            // Offline test: mocked responses are replayed per request sequence, so only asserting on the
+            // generated endpoint catches a sub folder that is read from its parent's url
+            using (var context = await TestCommon.Instance.GetContextWithoutInitializationAsync(TestCommon.TestSite))
+            {
+                var parentFolder = new Folder { PnPContext = context, Parent = context.Web };
+
+                var entityInfo = EntityManager.GetClassInfo<IFolder>(typeof(Folder), null, parent: parentFolder);
+
+                Assert.AreEqual("_api/Web/getFolderById('{Id}')", entityInfo.SharePointGet);
+                Assert.AreEqual("_api/Web/getFolderById('{Parent.Id}')/Folders", entityInfo.SharePointLinqGet);
+            }
+        }
+
+        #endregion
     }
 }
--- a/src/sdk/PnP.Core/Model/SharePoint/Core/Internal/Folder.cs
+++ b/src/sdk/PnP.Core/Model/SharePoint/Core/Internal/Folder.cs
@@ -19,11 +19,13 @@ namespace PnP.Core.Model.SharePoint
     /// Folder class, write your custom code here
     /// </summary>
     [SharePointType("SP.Folder", Target = typeof(Web), Uri = "_api/Web/getFolderById('{Id}')", LinqGet = "_api/Web/Folders")]
-    [SharePointType("SP.Folder", Target = typeof(Folder), Uri = "_api/Web/getFolderById('{Id}')", Get = "_api/Web/getFolderById('{Parent.Id}')", LinqGet = "_api/Web/getFolderById('{Parent.Id}')/Folders")]
+    [SharePointType("SP.Folder", Target = typeof(Folder), Uri = "_api/Web/getFolderById('{Id}')", LinqGet = "_api/Web/getFolderById('{Parent.Id}')/Folders")]
     [SharePointType("SP.Folder", Target = typeof(List), Uri = "_api/Web/Lists(guid'{Parent.Id}')/rootFolder", LinqGet = "_api/Web/Lists(guid'{Parent.Id}')/Folders")]
     [SharePointType("SP.Folder", Target = typeof(ListItem), Uri = "_api/Web/Lists(guid'{List.Id}')/Items({Parent.Id})/Folder")]
     internal sealed class Folder : BaseDataModel<IFolder>, IFolder
     {
+        private const string ParentFolderApiCall = "_api/Web/getFolderById('{Parent.Id}')";
+
         #region Construction
         public Folder()
         {
@@ -42,8 +44,12 @@ namespace PnP.Core.Model.SharePoint
                 //throw new ClientException(ErrorType.Unsupported, OnPCoreResources.Exception_Unsupported_AddingContentTypeToList);
                 //}
 
+                // A folder is added to its container, so when the container is a folder the add call has to
+                // target the parent folder. The Get of a folder points to the folder itself.
+                string containerApiCall = entity.Target == typeof(Folder) ? ParentFolderApiCall : entity.SharePointGet;
+
                 string encodedPath = WebUtility.UrlEncode(Name.Replace("'", "''").Replace("%20", " ")).Replace("+", "%20");
-                return new ApiCall($"{entity.SharePointGet}/Folders/AddUsingPath(decodedurl='{encodedPath}')", ApiType.SPORest);
+                return new ApiCall($"{containerApiCall}/Folders/AddUsingPath(decodedurl='{encodedPath}')", ApiType.SPORest);
             };
         }
         #endregion


Adam-it:
@svermaak we are happy you are eager to help us out but please do follow our [contribution guide](https://github.com/pnp/pnpcore/blob/dev/CONTRIBUTING.md) and [contributing best practices](https://pnp.github.io/pnpcore/index.html#best-practices-when-contributing).

The issue for which you opened a PR was not ready, it does not have the 'help wanted' label which means we did not agree on the fix direction and were not expecting anyone working on it.
Also you did not leave any comment in the issue do first get assigned before opening the PR. Please first get assigned to the PR and lets move the discussion first there to agree on the fix before we process this PR further

Adam-it:
Revisiting this PR in the context of #1586 (the issue is now confirmed as a bug and is being prepared for `help wanted`). I went through the code paths involved and I think the direction taken here is essentially correct. Writing down the verification so the analysis is not lost, whoever ends up picking this up.

### The diagnosis holds up

`EntityManager` falls back to `Uri` when `Get` is not set (`src/sdk/PnP.Core/Model/Base/EntityManager.cs`, ~line 75), and `QueryClient.BuildGetAPICallRestAsync` picks `entity.SharePointGet` for every non-LINQ read and `SharePointLinqGet` for collection queries (`src/sdk/PnP.Core/Services/Core/Query/QueryClient.cs:72`). So with

```csharp
[SharePointType("SP.Folder", Target = typeof(Folder), Uri = "_api/Web/getFolderById('{Id}')", Get = "_api/Web/getFolderById('{Parent.Id}')", ...)]
```

a `LoadAsync` on a folder-in-a-folder really is resolved against the parent folder's endpoint, while `.Folders` / `.Files` LINQ queries resolve `LinqGet` correctly. That is exactly the split the issue reports, and it also explains why it is not limited to collections.

Two extra pieces of evidence that the `Get` override was the anomaly rather than intentional:

- `Update` and `Delete` were never overridden on that mapping, so they already fell back to `Uri` (`{Id}`, the folder itself). A model whose write endpoint is itself but whose read endpoint is its parent is not a coherent mapping.
- The sibling `SP.File` mapping with `Target = typeof(Folder)` (`File.cs:19`) has `Uri = "_api/Web/getFileById('{Id}')"` and no `Get` override. Dropping `Get` here makes `SP.Folder` consistent with it.

### The `AddApiCallHandler` change

Reasonable, and it is the only consumer that depended on `Get` pointing at the container. Hardcoding the parent url for the folder target also matches what `FileCollection` already does (`_api/web/getFolderById('{Parent.Id}')/files/addusingpath(...)`, `FileCollection.cs:61,77,113`), so it is not a new pattern.

One small simplification worth considering: for the `Target = typeof(Folder)` mapping `LinqGet` is already `_api/Web/getFolderById('{Parent.Id}')/Folders`, so the add call for that target is just `$"{entity.SharePointLinqGet}/AddUsingPath(decodedurl='{encodedPath}')"`. That reproduces the previous url exactly and avoids carrying a duplicate literal that can drift from the attribute. (It only works for the folder target, `List`/`ListItem` still need `SharePointGet`, so the branch stays.)

### The test is the weak part

`SubFolderIsReadFromItsOwnEndpoint` asserts on `entityInfo.SharePointGet` / `SharePointLinqGet` string values, which is close to asserting the attribute back to itself and would not catch a regression introduced anywhere in the query pipeline.

The point about mocked tests being unable to catch this is correct as stated: `TestManager` keys mock files by test name plus request order (`{TestName}-{TestId}-{orderPrefix}.response.json`), so a wrong url replays the right payload. But there is a better existing pattern for exactly this kind of assertion: `PnP.Core.Test/Base/QueryTests.cs` uses `QueryClient.BuildGetAPICallAsync` via its `BuildModel` / `GetAPICallTestAsync` helpers and asserts the generated request url. Asserting the generated url for a sub folder load goes through the real code path instead of pinning attribute metadata, and would fail if the fallback logic or `QueryClient` changed.

### What still needs checking

The main behavioural risk of the switch is that reads now require the folder's own key (`UniqueId`) to be populated, where before only the container's id was needed. Worth confirming with live tests for at least: a folder obtained from `parentFolder.Folders`, a folder just created via `AddAsync` and then read back, `EnsureFolderAsync`, and the `List` / `ListItem` targets which should be untouched. Adding folder and file coverage under a nested folder to the live-recorded folder tests would give more confidence than the offline assertion.

Overall: the root cause identification looks right to me and the minimal `Get` removal is, in my opinion, the correct shape of fix rather than a workaround. @svermaak, sorry again about how the earlier exchange went, that was not the intent. If you would like to pick this up again once the issue gets the `help wanted` label, the changes here are a good starting point.

PR #1863, opened 2026-08-29
Scope ISharePointUser.Groups to the groups the user belongs to

### LINKED ISSUE #1630: Getting the groups of site user returns all the site groups instead of only the groups he's a member of

svermaak:
Fixes #1630

Thanks for the assignment @Adam-it, and for the analysis on #1839 that this builds on.

## The problem

`SP.Group` carried a single `SharePointType` attribute. `EntityInfo.SharePoint` returns the one mapping whenever `SharePointTargets.Count == 1`, regardless of parent, so `QueryClient.BuildGetAPICallRestAsync` resolved `SharePointLinqGet` to `_api/Web/SiteGroups` for every parent. `await user.Groups.ToListAsync()` therefore issued a site wide query with the user nowhere in it, and returned every group in the site.

## The fix

One mapping per parent. `Web` and `Site` keep the site wide enumeration, a `SharePointUser` parent enumerates `_api/Web/GetUserById({Parent.Id})/Groups`:

```csharp
[SharePointType("SP.Group", Target = typeof(Web), Uri = "...", LinqGet = "_api/Web/SiteGroups", Delete = "...")]
[SharePointType("SP.Group", Target = typeof(Site), Uri = "...", LinqGet = "_api/Web/SiteGroups", Delete = "...")]
[SharePointType("SP.Group", Target = typeof(SharePointUser), Uri = "...", LinqGet = "_api/Web/GetUserById({Parent.Id})/Groups", Delete = "...")]
```

`Uri`, `Get` and `Delete` stay site scoped on all three, only the collection enumeration is parent dependent.

The `Web` and `Site` attributes are required rather than cosmetic. Once `SharePointTargets.Count > 1`, resolution falls to `SharePointTargets.FirstOrDefault(p => p.Target == Target)` and returns null on no match, so leaving the original attribute untargeted would have null'd the mapping for `Web.SiteGroups`, `Web.Associated{Member,Owner,Visitor}Group` and `Site.HubSiteSynchronizableVisitorGroup`. I checked the parent surface and those, plus `SharePointUser.Groups`, are the complete set of parents for `ISharePointGroup` / `ISharePointGroupCollection` in the SDK, with nothing in `PnP.Core.Admin`.

## Scope

The endpoint returns **direct** SharePoint group membership. A group the user only reaches through an Entra ID security group that is itself a member is not expanded, which the issue also asked about. That is not something this endpoint can do and would need Graph, so I documented the limit on `ISharePointUser.Groups` rather than pretending it is covered. Happy to log the transitive part as a separate issue if you want it tracked.

## Test

`SharePointGroupLinqGetIsScopedToItsParent` asserts the **resolved** request url rather than reading the attributes back, mirroring `DataModelQueryService.AddToBatchAsync` so it also covers `{Parent.Id}` resolving off `Metadata[MetaDataRestId]` rather than the `Id` property. Mocked responses are replayed per request sequence rather than per url, so a wrongly scoped query would otherwise replay the right payload and the test would pass regardless.

Verified it actually catches the bug. Reverting only the model change fails it with the real buggy url:

```
Assert.IsTrue failed. Expected a user scoped query, got:
https://contoso.sharepoint.com/sites/pnpcoresdktestgroup/_api/Web/SiteGroups?$top=100
```

Full offline suite run: the 9 failures are all pre existing on `dev` on my machine (timezone, access token, `RepeatedCloneOffLineTest`, `SchedulePagePublish`, and one live call failing on a local TLS quirk). Nothing in the touched area.

## Evidence from a real tenant

Site with 6 groups, the user is genuinely a member of 2 of them. Same console app, same site, only the `PnP.Core.dll` differs. Step 3 also covers the other ways a user is reached.

**Before**

before

**After**

after

| Path | Before | After |
|---|---|---|
| `SiteUsers.FirstOrDefaultAsync(...)` then `user.Groups.ToListAsync()` | 6 groups | 2 groups |
| `web.EnsureUserAsync(upn)` | 6 groups | 2 groups |
| `web.GetUserByIdAsync(id)` | 6 groups | 2 groups |
| `group.Users` then that user | 6 groups | 2 groups |

Two notes on the remaining checks from your review:

- `web.CurrentUser.QueryProperties(p => p.Groups)` goes through `$expand`, not `LinqGet`, so it is unaffected. It is already covered by `UserTests.cs:59` and still passes.
- Paging is unchanged. Only the base `LinqGet` url differs, the `$top` and paging machinery around it is the same, and the generated request still carries the default `$top`.


Adam-it:
Thanks for opening up the PR. 
We will try to look into it ASAP

Adam-it (approved):
LGTM 🔥 
@svermaak awesome work. Only thing (as already stated in the PR description) this is a partial fix as it will return groups the user is direct memeber of and will not cover transitive Entra ID group membership, but it's good enough to merge. 
Lets state in the related issue it was partly fixed and perform a follow up PR if possible with the rest of the missing things

